PR #41, opened 2015-06-20
Use full company names for Manufacturer list

acolomb:
This is just a suggestion and will likely make the logs rather unreadable.

I think it would be best to extend the data structure to have ```name``` and ```shortName``` fields, where the latter defaults to the former when not used.  This would make it easy to not clobber the analyzer output, but have the genuine strings available if required.

What do others think? No strong preference on my side...


canboat:
I shortened the names from the manufacturer list on purpose so the company name is still recognisable but removed clutter such as " Corp." etc. (and to distinguish this effort from anything "officially" NMEA) so I'm leaning against this.

acolomb:
Alright, since nobody else commented I'm closing this to stay with the old names.

PR #54, opened 2016-01-23
Refactor analyzer, add pgn.c

jpilet:
I found it difficult to read analyzer.c, so I refactored it a little bit.

This change moves some pgn related code of analyser.c into pgn.c

It introduces a searchForPgn() function in pgn.h that has a sublinear complexity
to search through the list of known pgns.

I splitted printPgn, too.

The whole change should not change the behavior of "analyzer". No guarantee, though.
I would totally understand if you prefer the old structure.

### DIFF
--- a/analyzer/Makefile
+++ b/analyzer/Makefile
@@ -27,8 +27,8 @@ JSONFILE=pgns.json
 
 all: $(TARGETS)
 
-$(ANALYZER): analyzer.c analyzer.h pgn.h ../common/common.c ../common/common.h Makefile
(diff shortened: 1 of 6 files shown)


canboat:
Hi Julien,

I think this is very definitely something I'd like to pull. I've been itching to refactor analyzer for a long time as it is not very nice (but quite functional) code.

A code review of your pull showed that your new searchForPgn() function doesn't return the first PGN as the binary search can find something halfway down the array that happens to be the 2nd or further PGN for a particular PGN number. After search you need to back track to the first PGN that has the same PGN number in order to be compatible.

Can you implement this?

jpilet:
I added a unit test to make sure there is no need to backtrack to the first pgn. searchForPgn always return the first one.

I still need to resolve conflicts, though.

PR #70, opened 2016-08-26
Output BITFIELDS as an INTEGER

sbender9:
I was getting strange negative numbers for the Engine Discrete Status values. 

I'm not sure if this is the right fix, but makes everything work for me.

### DIFF
--- a/analyzer/analyzer.c
+++ b/analyzer/analyzer.c
@@ -1150,7 +1150,7 @@ static bool printNumber(char * fieldName, Field * field, uint8_t * data, size_t 
     else
     {
 
-      if (field->resolution == RES_INTEGER)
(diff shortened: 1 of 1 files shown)


canboat:
I implemented a better solution, imo.

They are now printed (example with json output):

{"timestamp":"2016-08-23T03:22:49.221Z","prio":2,"src":17,"dst":255,"pgn":127489,"description":"Engine Parameters, Dynamic","fields":{"Engine Instance":"Single Engine or Dual Engine Port","Oil pressure":0,"Temperature":120.55,"Alternator Potential":327.67,"Total Engine hours":554240,"Discrete Status 1": ["Over Temperature"],"Discrete Status 2": ["Warning Level 2"]}}

Note that the array will contain 0-n strings, with one string for every bit that is 1 in the value.

PR #91, opened 2017-08-21
add format detect  candump logfile in candump2analyzer

cmotelet:
I've added the format detect of  candump log file in candump2analyzer to be able to read file generated by candump with option -l.


canboat:
Thanks, looks good!

PR #95, opened 2017-09-24
updated PGN-65285

lfrancois:
Inverted `temperature instance` and `temperature source` to get the correct data


canboat:
You have not really inverted them, you have also made Temperature Source a 1 byte field. Although is this possible, this also means that the temperature itself, a 2 byte field, now starts half way through a byte. Are you really sure you are getting valid temperatures this way? 99.9% of all multiple byte fields start at a byte, so I think I have a better fix which works with the samples you provided.

You can reopen the issue if you disagree.

lfrancois:
It looks I was getting correct temperature, but I am all for your better fix :)

PR #98, opened 2017-12-04
Add fflush

zubenubi:
add fflush to end of while loop


zubenubi:
change suggested by @kees to overcome a lag noted in piping data from candump->candump2analyzer->analyzer

PR #102, opened 2018-01-15
 fix: errors when writing too much to fast to the NGT-1

sbender9:
I really hate calling sleep to fix a problem, but don't know else to fix this


canboat:
What would be even better is a rewrite where we do a select() to check if the NGT-1 is ready to write before we do any writes, but I guess this is good enough for now as it will only happen if you try to saturate the NGT1 with many many messages.

PR #109, opened 2018-01-24
fix: PGNInfos from object to array

tkurki:
There are several PGNs that are in pgns.xml several
times as PGNInfos. The old pgns.json had the PGNInfos
converted to properties of a holder object, keyed by
the pgn number. This was valid JSON, but later properties
overwrote earlier ones and they were not available.

This commit changes the PGNInfo holder to be an array,
that allows access to the duplicates. Code that uses
pgns.json must be adjusted to account for this change.


tkurki:
@sbender9 this work for you?

tkurki:
Oops, not quite correct, I'll get back to this.

tkurki:
Fixed.

sbender9:
this works perfectly

PR #120, opened 2018-08-21
fix: PGNInfos from object to array

tkurki:
There are several PGNs that are in pgns.xml several
times as PGNInfos. The old pgns.json had the PGNInfos
converted to properties of a holder object, keyed by
the pgn number. This was valid JSON, but later properties
overwrote earlier ones and they were not available.

This commit changes the PGNInfo holder to be an array,
that allows access to the duplicates. **Code that uses
pgns.json must be adjusted to account for this change**.

### DIFF
--- a/analyzer/pgns.json
+++ b/analyzer/pgns.json
@@ -2,9 +2,9 @@
     "Comment":"See https://github.com/canboat/canboat for the full source code",
     "CreatorCode":"Canboat NMEA2000 Analyzer",
     "License":"GPL v3",
-    "PGNs":{
+    "PGNs": [
     
-      "59392":{
+      {
         "PGN":59392,
         "Id":"isoAcknowledgement",
         "Description":"ISO Acknowledgement",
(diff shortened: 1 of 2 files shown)


tkurki:
I had done this earlier, but now that I think of this more: we could aim for publishing a `canboatdata`package in npm for use in JavaScript applications like canboatjs.

That implies that npm is available, which in practice means that we can run prepublish scripts, where we could have more sane tools for generating pgns.json than xslt, which I believe is no longer available out of the box on most platforms. 

With better tools we could make the structure of the file more usable out of the box, like an array valued property per pgn number.

Your thoughts @sbender9 @keesverruijt ?

sbender9:
It would definitely be nice to have this published. I can put this on my TODO list.

@keesverruijt  In the mean time, it would be great to if we could go ahead and merge this so that I can more easily update canboatjs.

keesverruijt:
yeah no problem

keesverruijt:
(sorry this took a while, I was sailing in the Lofoten and when I came back it had slipped my mind...)

PR #124, opened 2018-09-27
Add PGN 65409: Airmar Speed Pulse Count

mairas:
I added one missing Airmar DST200 proprietary PGN. This one would be essential in generating calibration tables for their speed sensor.


mairas:
Ping?

PR #126, opened 2018-10-16
analyzer: add support for Yacht Digital YDWG-02

saxofon:
Hi,

Started to play a bit with this software. Adding support for another interface to the nmea2k bus.
So far only additions to analyzer and tested with a fairly empty nmea2000 bus, Raymarine/Airmar DST800 triducer only.
More testing to come and hopefully further integrations.

Signed-off-by: Per Hallsmark <per@hallsmark.se>


keesverruijt:
I would like you to make the format detection slightly more robust, there is an upcoming new device that will interfere with the detection you are doing.

E.g. instead of 

    if ( (msg[13] == 'R') || (msg[13] == 'T') )

for checking 

    21:55:38.937 R 09F50323 FF 00 00 FF FF 00 FF FF

I suggest you do something like 

    if ((sscanf(msg, "%d:%d:%d.%d %c %08X ", &a, &b, &c, &d, &e, &f) == 6)

saxofon:
No problem, commit updated.

PR #135, opened 2018-11-11
Added name & Id to PGN field w/ banks.

jonross09:
Just added field names. Makes my json & xml parser blow up seeing empty fields. Feel free to change the name to your liking.


keesverruijt:
See issue #136. Your pull request uses names that are not used for other reserved fields, so using a different name ("Reserved").

Issue #129, opened 2018-10-29
AIS encoder/!AIVDM in n2kd


free-x:
Hi,

whether this will eventually be implemented
``` 
// This does not work yet. It needs to be encoded to format like !AIVDM,1,1,,B,177KQJ5000G?tO`K>RA1wUbN0TKH,0*5C - from
  // http://catb.org/gpsd/AIVDM.html nmea0183CreateMessage(msg183, src, "PVDM,%d%d%d,%s,%s,%s,%s,%s,%.3f,%s,%s", utc_time->tm_hour,
  // utc_time->tm_min, utc_time->tm_sec, mmsiString, latString, lonString, sogString, headingString, cog, rateOfTurnString,
// navigationStatusString);
```
i have already searched for the AIS-encoder implementation in C. Found this only in gpsd(http://git.savannah.gnu.org/cgit/gpsd.git/tree/pseudoais.c) . The question is whether it can be used without further ado, because gpsd was developed under BSD license but "canboat" is under GPL-3

Regards
free-x

keesverruijt:
The AIS messages contain binary data, encoded using a variation of Base64 encoding. I am in the process of adding base64 encoding/decoding, it wouldn't be hard to add the mapping used by AIS. This will hit the master branch within the next 2 weeks.

The gpsd code doesn't fit our use case anyway, but it will help establishing the bit patterns in the AIS ITU format.

svdrummer:
Would this be needed in CANboat. To transmit in AIS, all owners need a
licence number to be entered into the AIS tranciever at time of
installation. If you are able to receive AIS, that data is already encoded
and formatted.

keesverruijt:
@svdrummer the use case is where you have an AIS connected only to NMEA 2000, not 0183, and you have a program that does not handle Signal K or CANboat style NMEA 2000 data, e.g. it can only process NMEA0183 style `!AIVDM` messages.

free-x:
#156 

hopefully it will be merged soon

larsujensen:
As a short time solution I suppose it is possible to clone my fork of canboat today, checkout the branch ais and build it. Just do not rely on me keeping my fork updated with canboat. And keep in mind that it is a "prototype", take it as it is.

free-x:
Hi @larsujensen,
I want to thank you for your great job. I checked out your "ais" branch and played with known samples. Only candumpSample3.txt contains AIS information.

```
[...]

Issue #92, opened 2017-08-27
Improve handling of variable length strings, especially in JSON and XML descriptions


ttlappalainen:
Definition for 126998 is wrong. It has only fields:
1 Installation Description, Field 1
2 Installation Description, Field 2
3 Manufacturer Information, Field 3
All fields are variable length strings so that there is:
<len><0x01><string data>, where len is length(<string data>)+2. So this PGN message can be:
<0x02><0x01><0x03><0x01><"C"><0x05><0x01><"CAT">

Also on PGN 129802 field 7 "Safety Related Text" is constructed as above - not fixed length field. And possibly on other PGNs containing text data.

canboat:
Can you upload some raw PGN strings, in a format that analyzer understands?
I have no sources providing these PGNs and since this is all produced by reverse-engineering ... 

There are (at least) four ways of representing strings in N2K, so with the 'have never seen' definitions I had to guess, and I guess I guessed wrong here :-)

ttlappalainen:
Have you ever checked my NMEA Simulator (http://www.kave.fi/Apps/NMEA-Simulator/NMEA-Simulator.7z - unfortunately only Windows version available). It sends all data to selected port and also configuration information (126998) in Actisense format. I normally test and play with simulator either by null modem cable (com0com driver or physical) or through real N2k bus by using my NMEA 2000 library sample ActisenseSender.

ttlappalainen:
I have actually wondered how many designer they have needed to create mesh like N2k. But could you point me other ways for strings. I know only these two with fixed field like on product information (126996) and variable field on some others like 126998 and 129802.

canboat:
The CANBoat analyzer has:

RES_ASCII (fixed field length, no byte count)
RES_STRING (starts with 1 byte length byte)
RES_STRINGLZ (starts with 1 byte length byte and terminated by zero byte)
RES_STRINGLAU (starts with 1 byte length byte and 1 byte ASCII/UNICODE byte)

The RES_STRINGLZ variation is used by Fusion for the old company specific PGNs. Not sure about the new generic N2K standard audio PGNs -- I have not seen them so have not yet added them.

The RES_STRINGLAU variation is used by the new N2K PGNs since for DSC. The analyzer doesn't yet properly handle the UTF16 content variation. (Why in the world you would choose UTF16 instead of UTF8 when coming up with this in 2014-2015 is really beyond me, but okay.)

ttlappalainen:
[...]

PR #156, opened 2019-03-18
Ais

larsujensen:
The branch ais creates nmea0183 ais sentences for the PGNs listed in the commit message. The code has been installed in a boat with an em-trak b100 class B ais, producing ais objects on the chart and radar in opencpn on a Raspberry Pi. I believe the list of implemented ais messages are rather complete for use for navigation in a leisure boat, and to the best of my knowledge it covers all the pgns sent by an em-trak b100. Thus, to me this project is finished, except for maintenance, correcting found bugs and changes initiated by changes in the rest of canboat, it is tied rather hard to the output format of analyzer.

Testing is a problem. Lack of open documentation of PGNs makes it impossible to create reliable artificial test data. Therefore test data must come from real ais generated pgns. The problem is that some types of ais messages are rare in real life. The output, the ais sentences, is an encoded bit vector. Verifying the vector requires decoding. Once again, lacking a formal open documentation of the format, this decoding can not be based on the same description used in creating the sentences.
I have used a real traffic candump file, passed it through analyzer, thus generating an input json test data file for n2kd. This file contains ais messages for common block (1,2,3), 4, 5, and 18. I have copied the generated sentences to a web based decoder and compared with the content of the json file. I have also compared some of the result of message 5, which is ship static and voyage related data, with actual ship data found on internet, thus in some sense also verifying analyzer.
At least one more type of message has been decoded in the boat, 19 and/or 24, superficially looking good but not verified.

There are untested ais messages due to lack of pgns. I have looked though the code, and essentially all branches in the code are executed by the tested messages, exceptions are "exeptions" like missing fields in the json file. But there are two fairly simple functions not used by the tested messages. I have tested these by a small external test program to at least verify that the do as intended. This does not verify that the intention is without flaws.

The code is to a large extent "table driven". Although the instructions are verified by some tests this does not verifies all of the tables. But this is an issue shared with the rest of canboat. There are some issues of canboat, generally documented as comments in the code, that should be discussed.

Ais is a navigational tool, and as such it is important that essential information is not false. In my opinion the most essential information is position, possibly followed by COG. Longitude and latitude are given by signed integers. The practical testing in the boat in the harbor of Stockholm gave a pretty good confidence in position. But this is for positive integers. In the test file there where westerly positions, i.e. negative longitude, confirmed consistent with the json test data. As longitude and latitude essentially shares the same code I am fairly confident that it is correct, but it is not proven.

### DIFF
--- a/n2kd/gps_ais.c
+++ b/n2kd/gps_ais.c
@@ -530,11 +530,19 @@ long int aisFloat(const char *msg, const char *fieldName)
   int      sign;
   long int result;
 
-  floatParam paramRange[] = {// pgns.json says rad/s, but output indicates deg/s
-                             {p0, "Rate of Turn", -127, 127, -128, SECONDS_PER_MINUTE},
+  floatParam paramRange[] = {{/* Min and max values of RoT are -127 and 127, but these values have
+                                 special meaning, which is currently not present in canboats
+                                 definition of the parameter. We therefore defaults if we
+                                 get values with magnitude above 126.
+                              */
+                              p0,
+                              "Rate of Turn",
+                              -126,
+                              126,
+                              -128,
+                              SECONDS_PER_MINUTE},
                              {p1, "SOG", 0, 1022, 1023, KNOTS_IN_MS * SOG_MULTIPLICATOR},
                              {p2, "COG", 0, 3599, 3600, COG_MULTIPLICATOR},
-                             // pgns.json says rad, but output indicates degrees
                              {p3, "Heading", 0, 359, 511, NO_MULTIPLICATOR},
                              {p4, "Longitude", -108000000, 108000000, 0x6791AC0, LON_MULTIPLICATOR},
                              {p5, "Latitude", -54000000, 54000000, 0x3412140, LAT_MULTIPLICATOR},
@@ -543,7 +551,6 @@ long int aisFloat(const char *msg, const char *fieldName)
                              {p8, "Position reference from Starboard", 0, 630, 0, SIZE_MULTIPLICATOR},
                              {p9, "Position reference from Bow", 0, 5110, 0, SIZE_MULTIPLICATOR},
                              {p10, "Draft", 0, 255, 0, DRAUGHT_MULTIPLICATOR},
-                             // Assumed degrees as in "Heading", despite rads
                              {p11, "True Heading", 0, 359, 511, NO_MULTIPLICATOR},
                              {p12, "Length/Diameter", 0, 10220, 0, SIZE_MULTIPLICATOR},
                              {p13, "Beam/Diameter", 0, 1260, 0, SIZE_MULTIPLICATOR},
@@ -570,14 +577,7 @@ long int aisFloat(const char *msg, const char *fieldName)
   switch (paramRange[i].hash)
   {
     case p0:
-      result = ROT_MULTIPLICATOR * sqrt(value) + 0.5;
-      break;
-    case p2:
-      result = ((long int) (value + 0.5)) % 3600;
-      break;
-    case p3:
-    case p11:
-      result = ((long int) (value + 0.5)) % 360;
+      result = (long int) (ROT_MULTIPLICATOR * sqrt(value) + 0.5);
       break;
     case p10:
       result = (long int) (value + 0.9);
@@ -587,10 +587,10 @@ long int aisFloat(const char *msg, const char *fieldName)
       break;
   }
   result *= sign;
-  if (result < paramRange[i].min)
-    return paramRange[i].min;
-  if (result > paramRange[i].max && result != paramRange[i].defValue)
-    return paramRange[i].max;
+  // Make shure values are valid
+  if (result != paramRange[i].defValue)
+    if (result < paramRange[i].min || result > paramRange[i].max)
+      result = paramRange[i].defValue;
   return result;
 }
 
--- a/n2kd/gps_ais.c
+++ b/n2kd/gps_ais.c
@@ -186,64 +186,1120 @@ void nmea0183GLL(StringBuffer *msg183, int src, const char *msg)
 }
 
 /*
-{"timestamp":"2015-12-14T22:06:21.609Z","prio":4,"src":0,"dst":255,"pgn":129038,"description":"AIS Class A Position
-Report","fields":{"Message ID":1,"Repeat Indicator":"Initial","User
-ID":224593000,"Longitude":-5.4368332,"Latitude":36.1309824,"Position Accuracy":"Low","RAIM":"not in use","Time
-Stamp":"21","COG":165.0,"SOG":0.00,"Communication State":"33576","AIS Transceiver information":"Channel A VDL
-reception","Heading":51.0,"Rate of Turn":0.11,"Nav Status":"Under way using engine","Regional Application":1}}
-{"timestamp":"2015-12-14T22:06:21.662Z","prio":4,"src":0,"dst":255,"pgn":129039,"description":"AIS Class B Position
-Report","fields":{"Message ID":18,"Repeat Indicator":"Initial","User
-ID":235015519,"Longitude":-5.3561452,"Latitude":36.1571296,"Position Accuracy":"Low","RAIM":"not in use","Time
-Stamp":"5","COG":0.0,"SOG":0.00,"Communication State":"393222","AIS Transceiver information":"Own information not
-broadcast","Heading":31.0,"Unit type":"CS","Integrated Display":"No","DSC":"Yes","Band":"entire marine band","Can handle Msg
-22":"Yes","AIS mode":"Autonomous","AIS communication state":"ITDMA"}} !AIVDM,1,1,,B,177KQJ5000G?tO`K>RA1wUbN0TKH,0*5C - from
-http://catb.org/gpsd/AIVDM.html
-
-AIVDM - Automatic Information System (AIS) position reports from other vessels - from http://opencpn.org/ocpn/nmea_sentences
-1. Time (UTC)
-2. MMSI Number
-3. Latitude
-4. Longitude
-5. Speed Knots
-6. Heading
-7. Course over Ground
-8. Rate of turn
-9. Navigation status
-
-These are not rate limited, as each one is for a different vessel.
+=== AIS VDM/VDO ==
 */
 
-void nmea0183AIVDM(StringBuffer *msg183, int src, const char *msg)
-{
-  struct tm *utc_time;
-  time_t     current_time;
+typedef struct {
+  char bitVector[226]; // Corresponds to 300 6-bit characters. More than currently needed.
+  int pos;
+} aisVector;
 
-  char   mmsiString[MMSI_LENGTH];
-  char   latString[LAT_LENGTH];
-  char   lonString[LON_LENGTH];
-  char   sogString[SPEED_LENGTH];
-  char   headingString[ANGLE_LENGTH];
-  char   cogString[ANGLE_LENGTH];
-  char   rateOfTurnString[ANGLE_LENGTH];
-  char   navigationStatusString[OTHER_LENGTH];
-  double cog = 0;
+/*
+Adds integer values to the packed 6 bit byte ais vector.
+*/
+static int addAisInt(long int value, int len, aisVector *payload) {
+  int i, k;
+  unsigned char nibble;
 
-  current_time = time(NULL);
-  utc_time     = gmtime(&current_time);
+  // Break if overflow, does not treat signed positive value properly
+  if (value >= (long int)(1 << len) || value < -(long int)(1 << (len - 1))) return 0;
 
-  getJSONValue(msg, "User ID", mmsiString, sizeof(mmsiString));
-  getJSONValue(msg, "Latitude", latString, sizeof(latString));
-  getJSONValue(msg, "Longitude", lonString, sizeof(lonString));
-  getJSONValue(msg, "SOG", sogString, sizeof(sogString));
-  getJSONValue(msg, "COG", cogString, sizeof(cogString));
-  getJSONValue(msg, "Heading", headingString, sizeof(headingString));
-  getJSONValue(msg, "Rate of Turn", rateOfTurnString, sizeof(rateOfTurnString));
-  getJSONValue(msg, "Nav Status", navigationStatusString, sizeof(navigationStatusString));
+  // AIS uses big endian. Thus most significant bits shall be stored first.
+  while (len > 0) {
+    i = payload->pos / 8;
+    k = 8 - payload->pos % 8;
+    payload->bitVector[i] &= ~(unsigned char)((1 << k) - 1);
+    if (len >= k) {
+      nibble = ((1 << k) - 1) & (value >> (len - k));
+      payload->pos += k;
+    }
+    else {
+      nibble =  ((1 << k) - 1) & (value << (k - len));
+      payload->pos += len;
+    }
+    payload->bitVector[i] |= nibble;
+    len -= k;
+  }
+  return 1;
+}
 
-  cog = strtod(cogString, 0);
+/*
+Adds ascii strings to the packed 6 bit byte ais vector.
+*/
+static int addAisString(char* string, int len, aisVector *payload) {
+  int i, k;
+  char *j;
+  unsigned char nextchar;
 
-  // This does not work yet. It needs to be encoded to format like !AIVDM,1,1,,B,177KQJ5000G?tO`K>RA1wUbN0TKH,0*5C - from
-  // http://catb.org/gpsd/AIVDM.html nmea0183CreateMessage(msg183, src, "PVDM,%d%d%d,%s,%s,%s,%s,%s,%.3f,%s,%s", utc_time->tm_hour,
-  // utc_time->tm_min, utc_time->tm_sec, mmsiString, latString, lonString, sogString, headingString, cog, rateOfTurnString,
-  // navigationStatusString);
+  if (string == NULL) return 0;
+
+  while (len >= 6) {
+    // Encode current char to 6-bit ascii
+    // 32 -> 63 => 32 -> 63, 64 -> 95 => 0 -> 31
+    nextchar = *string;
+    if (nextchar == '\0') nextchar = 0;
+    else if (nextchar < 32 || nextchar > 95) nextchar = 32;
+    else if (nextchar >= 64) nextchar -= 64;
+    // Add 6-bit ascii char to payload
+    i = payload->pos / 8;
+    k = 8 - payload->pos % 8;
+    payload->bitVector[i] &= ~(unsigned char)((1 << k) - 1);
+    if (k >= 6) {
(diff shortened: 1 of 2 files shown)


keesverruijt:
Hi Lars,

I had a quick browse, initial impression is very favourable.

Would you mind running these through `clang-format`? It should pick up the `.clang-format` settings file in the project directory and apply the same formatting as we have in the rest of the source code.

larsujensen:
As can be seen from above I have made some minor corrections. I also added PGN 129801 "AIS Addressed Safety Related Message".
I have so far tested with the following PGNs:
PGN 129038 "Class A position report"
PGN 129039 "AIS Class B Position Report"
PGN 129793 "AIS UTC and Date Report"
PGN 129794 "AIS Class A Static and Voyage Related Data"
PGN 129041 "AIS Aids to Navigation (AtoN) Report"
PGN 129809 "AIS Class B "CS" Static Data Report, Part A"
PGN 129810 "AIS Class B "CS" Static Data Report, Part B"
Naturally the tests are not exhaustive. 
The following PGNs are untested as I have no test data for these:
PGN 129040 "AIS Class B Extended Position Report"
PGN 129798 "AIS SAR Aircraft Position Report"
PGN 129801 "AIS Addressed Safety Related Message"
PGN 129802 "AIS Safety Related Broadcast Message"

tkurki:
How are you testing? You don't happen to have nice input => expected output data that could be added here and possibly used also elsewhere, mainly in https://github.com/SignalK/signalk-to-nmea2000?

larsujensen:
Unfortunately no, I do not have such  a test suite. I would have been nice for regression test. But I tried to explain the problem with testing when I made the pull request.

Basically I produce a test file for n2kd by running live ais traffic recorded by candump through analyzer. The file I get is a json file, thus both readable and searchable. This file can be passed through n2kd with the debug option, thus producing nmea0183 sentences. I have used the ais decoder at https://www.maritec.co.za/tools/aisvdmvdodecoding/ to decode the sentences. The decoder writes out the values for the different fields. I have manually compared these values with the input data in the json file, where necessary manually doing the required conversions. (Notice the decoder does not produce expected position data, expected format and value, in message 1,2,3 (common block), but it seems ok in all other types of messages.)

All the tested PGNs I mentioned above are in the candump file I uploaded in #issue 129. I believe PGN 129040 is more or less depreciated, so what is lacking in the dump is safety related messages.
I have created a more manageable test file by copy and past of selected pgns from the rather huge json file produced by passing the candump file through analyzer.

There is a problem using n2kd debug option this way, it produces a lot of duplicates, I believe because the logDebug call in nmea0183CreateMessage(...) in file nmea0183.c prints all sentences in msg183 each time a new sentence is added. This could of course be solved. But when I wrote the code I had a small test program reading from the test file and running gps_ais.c in a debugger and replacing nmea0183CreateMessage(...) by a function directly printing created sentences. Thus no duplicates and quite easy to copy to the maritec decoder. Anyhow, I don't think you are using n2kd in SignalK.

Testing this way does not test the whole chain, as it assumes analyzer produces valid data. We have it installed in the boat running opencpn. It gives much credibility to the data produced seeing moored ships moored at the quays, ships sailing sailing forward where ships should sail, seeing e.g. Mariella 
with destination Mariehamn and ETA in the middle of the coming night (surely both of us have seen the ferries from Stockholm arriving in Mariehamn in the middle of the night). And seeing both real and virtual AtoN in reasonable positions. And more.

tkurki:
>I have created a more manageable test file by copy and past of selected pgns from the rather huge json file produced by passing the candump file through analyzer

What I was after here was just this: some _selected pgns_ and the respective, expected 0183 output that you have verified manually with an external tool. Now if (when!) we refactor the code somebody could use that data to verify that things are still working as expected, without using the external tool.

Mariehamn: yes, been there, done that, probably will again in the coming summer ;-)

larsujensen:
I can see what you are getting at. But the tests I have done is so far from exhaustive that I can not see that it will work with any confidence.
But surely, in principle I suppose it would be possible to use my code as some form of reference implementation. I think it would be wise to have it better verified first. And strive for a decently high coverage in the tests.

larsujensen:
I have in a commit made a change in Rate of Turn in  "Class A position report". PGN 129038 says rad/s, and data suggests this is not possible. I have come to the conclusion that it could be deg/s, and changed accordingly. See the change as experimental.

But, I would guess there are more to the Rate of Turn parameter in PGN 129038 than just a real value. Following is what it is supposed to reflect in the 8 bit parameter in the over the air message:
0 to +126 = turning right at up to 708 deg per min or higher
0 to -126 = turning left at up to 708 deg per min or higher Values between 0 and 708 deg per min coded by ROTAIS = 4.733 SQRT(ROTsensor) degrees per min
where ROTsensor is the Rate of Turn as input by an external Rate of Turn Indicator (TI). ROTAIS is rounded to the nearest integer value.
+127 = turning right at more than 5 deg per 30 s (No TI available)
-127 = turning left at more than 5 deg per 30 s (No TI available)
-128 (80 hex) indicates no turn information available (default).
ROT data should not be derived from COG information.

I would expect something corresponding to the values +127 and -127 be present in the PGN parameter.

Thus "Rate of Turn" possibly has similarities with "Time Stamp":
UTC second when the report was generated by the electronic position system (EPFS) (0-59, or 60 if time stamp is not available, which should also be the default value, or 61 if positioning system is in manual input mode, or 62 if electronic position fixing system operates in estimated (dead reckoning) mode, or 63 if the positioning system is inoperative)

To the best of my knowledge canboat implements Time Stamp as above.

keesverruijt:
*ALL* data fields in NMEA2000 are in SI units, with all angles and angular changes expressed in radians (rad) and radians/second. All data fields that require a fractional value are expressed not in floating point but using a decimal fraction.

According to my data angles are expressed in thousands of rad, e.g. 0.001 rad. The rotation fields in NMEA2000 use either 1/32 of a thousand, or 1/32 of a millionth rad/s(!) This matches up with what I see in the field (compass rotations).

This causes main pain points for the AIS messages, as their over-the-air format uses completely different units. For angular changes (e.g. ROT) these will indeed be in deg/s.

There are MANY ais'es especially older ones that encode the NMEA2000 messages wrong. For instance the SRT1 devices forgot to encode the "special value" for "No Heading Present" which in the binary format was 2^9-1 = 511, as the corresponding "special value" (0xffff) in the NMEA pgns, so they would just transmit "511 degrees" which was, modulo 360, 151... So all Class A vessels that did not have a compass plugged in would show at this particular angle on the chart, with such AISes.

woobagooba:
Kees.  These problem AIS units of which you speak.  Are they NMEA certified?

keesverruijt:
Yes, they were. Long replaced though, 2 generations ago this was.

Note that I can understand the problem: you have 2 different “ranges” of values for various different datatypes, and need to really get all the outliers correct. The developers didn’t find it since Class A transmitting ships usually have a mandatory compass input. It was inland shipping at the time that didn’t need that, so there was a lot of that here in the Netherlands, but apparently none or not enough where those developers were.

It did result in me getting a free 2nd gen test AIS …

Kees

larsujensen:
I think we are in for an interesting discussion. I want to lay some ground for this discussion first.

1, AIS is covered by specific PGNs. I am only talking about these PGNs. I to thought that NMEA2000 fields are in SI units, and I am not questioning that for any PGNs other than the AIS PGNs. So when you have a compass unit supplying radians I am in no way questioning that.

2, Our AIS is about three years old, but I have candumps from two different sources, different AISes, indicating the same thing in what will follow.

3, part of what I am saying has nothing to do with an AIS, it is lack of information in the definition of PGN 129038 in canboat.

4, I am really not talking about what is on the bus, but what is coming out of analyzer.

5. And finally something to think about. NMEA 0183 $ sentences is kind of an application protocol, with comma separated arguments, and a header indicating type of sentence and thus the interpretation of the parameters. Not so for AIS over NMEA 0183. AIS is an over the air application protocol, conveying information between radio units. NMEA 0183 is just used as a transport protocol for the AIS message over the wire, the complete ais message being in a single NMEA 0183 parameter. And the interpretation of the ais bit vector is not given by the head of the sentence but by the first 6 bits of the bit vector. The types of arguments in $ sentences are a completely different issue from the types of arguments i ais, despite ais being carried over NMEA 0183, they are in two different name spaces.

Suppose this concept is carried over to NMEA2000. Basically using NMEA2000 as a transport protocol. A priori with a separate ais name space for the arguments of the ais pgns. They have in N2k taken the type of message (the first 6 bits) and giving each type a separate PGN. (And also used an extra sub type bit in message 24 to give each sub type its own pgn.)  The over the air ais protocol is somewhat compressed, this compression is also removed in N2k, making it easier to process. But essentially, field for field there is a one-to-one correspondence between the over the air ais and the pgn. So essentially I am suggesting that the ais name space, i.e. the data types used in over the air protocol is valid also in n2k, but they are stored differently as they are not compressed (and in some cases they are also mangled, "modernized"). And when appropriate reusing data types from the NMEA2000 "application" protocol.

So far speculations. Lets move on to observations.
There is course information, COG, in pgns 129038 and 129039 coming out of analyzer. The values I have seen are in the range 0 - 359, very consistent with degrees, as defined in ais, but far outside the range of radians, which reasonably should be limited to < 2pi. Supplying this information interpreted as degrees to opencpn gives a green arrow on the chart, and I suppose opencpn uses COG to give the direction the arrow point in. Now, if that very same ship is moving, new positions arrive moving the ship on the chart. If the ship, as observed, moves in the same direction as the arrow point in I think it is a very good hint that interpreting ais cog as given in degrees is a good interpretation. And I think it is a fare flung idea that a modern well known ais is supplying wrong data on NMEA2000. 

Now back to AIS Rate of Turn, as defined in the ais protocol. It carries information that reasonably has no correspondence in a rate of turn from a compass. This information supplied over the air is no where to be found in the definition of PGN 129038 in canboat. I suggest that the definition in canboat is wrong for this particular pgn, and that it is, due to the general one-to-one correspondence i mentioned, the definition of the Rate of Turn field that is wrong. I.e. PGN129038 AIS Rate of Turn is a data type completely different from normal NMEA2000 Rate of Turn. But I may be wrong.

Once again, lets move to observation.
I can follow a ship in my data from analyzer. It changes RoT, and turns at most with a value 1.25 coming out of analyzer. Now, I doubt any ship with an AIS Class A transponder is able to turn 72 degrees per second. But I do not know how the ship turned in real life as I did not pay attention to it when i recorded the data. I admit I am a bit confused. If I read out the value 1.25 and it says it is rad/s then I assume analyzer has scaled it back to rad/s, which generally would be a number with many zeroes in the beginning. Am I getting something wrong here?

However, I have dived into the data, turning on most of what I could find of debug information in analyzer. As far as I could see a little endian hex value processed by analyzer into 1.02 was in decimal notation 567. If I divide this by 1000*32, as you indicate, I get 0,01771875. How the heck do you arrive at 1.02? However converting 0,01771875 rad/s to deg/s gives rounded up 1.02 deg/s. And 1.02 deg/s is 61 deg/min, a much more sensible value.
And pasting all the generated sentences with the degree interpretation into the decoder shows a much more reasonable turn of the ship.
And by the way the field is a 16 bit signed value. The scaling you mention uses 15 bits, so to me it looks as if 1.25 would be rad/s it would have been an overflow.

I am looking forward to see your comments on this!

larsujensen:
It took me a long time to write my comment above, so I have not seen your earlier comment until after I sent my own comment. Just for information.

keesverruijt:
Simple, analyzer converts values back to human readable values unless you use the `-si` flag. 

       if (field->resolution == RES_RADIANS)
        {
          units = "rad";
          if (!showSI)
          {
            a *= RadianToDegree;
            precision -= 3;
            units = "deg";
          }
        }
        else if (field->resolution == RES_ROTATION || field->resolution == RES_HIRES_ROTATION)
        {
          units = "rad/s";
          if (!showSI)
          {
            a *= RadianToDegree;
            precision -= 3;
            units = "deg/s";
          }
        }

larsujensen:
We are both right, I guess. When I look at what I wrote about the rate of turn value it seems to be rad/s in the pgn, but analyzer converts it to  deg/s, in this magical step from 0.01771875 to 1.02. With candumps timing of can frames RoT values looks consistent with COG values from "adjacent" pgns from the turning ship. 

If I convince you that analyzer is delivering deg for COG and deg/s for RoT I have just another little comment. The resolution from analyzer with just two decimals is a bit low, giving a minimal RoT of 0,01 deg/s. AIS has a minimal RoT of 0.0007 deg/s, and the pgn has a minimal RoT of 0.00003125 if I get you right. At least so it presently looks to me.

The magical step I leave to you!

larsujensen:
And now I haven't read your last comment again. So I am missing a flag, both in my test and in the configuration in the boat! I suppose this will also solve my complaints about precision. I have run my candump through analyzer with the si flag, and yes it works.

This is a nasty one for amateur users as me. You are as far as I can tell not mentioning this flag in the wiki. We have noticed last summer that SOG was wrong in opencpn, that it seemingly didn't convert properly between m/s and knots. Thought it was a bug.

But I am of course using -json, is it a good idea to have this output affected by such a flag? Or am I again missing something.

larsujensen:
Kees, I am a bit stunned. There are many flags in analyzer. With analyzer piping data into n2kd, which flags, in addition to -json, are supposed to be set, and which flags are supposed not to be set, for nmea0183 to work properly? Not only my code, but for all generated nmea0183 sentences? I have learned the hard way that the -si flag should be set, but how about other flags?

With the -json flag set, do flags like e.g. -camel or -upper-camel or -geo affect the format of the message parsed by getJSONValue? I.e. do the keys, the field names, change, and do the values, what comes out of getJSONValue, change?

I am not certain any more it was a good idea of me to make this pull request. If people, like me, use ais and sets the flags wrong ais may produce valid sentences but with false data. And I suppose this goes for other nmea0183 sentences as well.

larsujensen:
Kees, I have been thinking, or at least trying to. If ais is eventually going to be merged with the master branch of canboat, ais must use the same settings of flags in analyzer as is required for present none ais nmea0183 sentences. nmea0183 uses degrees. Looking at Heading in nmea0183.c I can not see any conversion from radians to degrees. So I assume it should be used without the  -si flag.

It would be fairly easy to change the other sentences to use radians in stead of degrees. But I assume the output of the other serves of n2kd are also affected by the -si flag. Considering this it seems to me that for backward compatibility we must stick with no -si flag.

I have made the change in ais to use the -si flag and was going to install it in the boat in the weekend to see that it behaved as expected, but I've put this on hold.

I would like to hear your opinion about this.

If we proceed without the -si flag i would suggest to look over if the precision of rate of turn could be increased with maybe a couple of decimals. Not necessary but nice. And I think it could be done without seriously harming the backward compatibility.

keesverruijt:
Take a look at the `nmea0183_units` branch. I now pass the `-si` flag on to n2kd, which then converts units back to NMEA0183 units.

larsujensen:
I have looked at the branch, and I think it would be trivial to include this in ais. But I have branched from the master so some merging needs to be done first. I think I should stick to not using -si flag until this is solved. This way ais is usable.

larsujensen:
Regarding rate of turn. Now that it is settled that it is really deg/s without the -si flag I feel fairly confident that the code provides correct values. I have done a little bit of tracing of a ship in exel based on the figures coming of analyzer, and with reasonable confidence it seems that the turn rates decently matches COG. And the turn rates of the decoded ais sentences matches the turn rates from the pgn.
I solve the remaining issue with the two special values,
+127 = turning right at more than 5 deg per 30 s (No TI available)
-127 = turning left at more than 5 deg per 30 s (No TI available)
by limiting the accepted turn rates to 0 -  +- 126 after conversion from pgn values. Turn rates outside this range results in -128, i.e. turn rate not available. I plan to file an issue as I think it is worth a warning to others to not using the Rate of Turn parameter from pgn 129038 without a range check.

larsujensen:
Teppo brought up the issue of automated testing, and I lack a facility for easy regression testing. Something badly needs to be done, because the minor changes I make in the code is trivial, but assuring everything works after each build is a burden.
Being a newcomer in open source projects at github I could need some advise. And the first question popping up is, when changes are made in canboat, say in analyzer, how is testing performed?

larsujensen:
As you may see I have made the restriction in Rate of Turn I mentioned above. Ais should currently be used without -si flag in analyzer.

I know of no errors in produced sentences not related to limitations in the output from analyzer. I have filed a couple of comments in issues, two in new issues, and one in issue #92, and plan to file a couple more.

larsujensen:
I'm uploading a zip archive with the "json_reg_test_file" I am using as input to n2kd as well as a python script, "aisfilter_unique", which is a filter that filters out ais sentences of the debug output from n2kd and removes duplicate ais messages. Notice messages, not sentences, the distinction is important. The filter preserves the order of ais messages. Do not expect the filter to be fool prof.

Data in the json file is mostly from the candump file I previously uploaded in issue #129, but some are from candumpSample3.txt in canboat. For some reason the first pgn seems not to come out of n2kd debug so I have duplicated this first pgn.

It should now be possible to do something like the following:

cat json_reg_test_file | n2kd -d 2>&1 | ./aisfilter_unique > outfile

and get a result in outfile which is equivalent to the content of file "json_reg_test_file_output", which is also included in the zip archive.
The output can then be pasted into https://www.maritec.co.za/tools/aisvdmvdodecoding/ to verify the result.

I think this is a principal step on the road to the test Teppo asked for, testing against json_reg_test_file_output, although the coverage could be higher. Also one step remains to be done in the test. The test input file should be a can frame file, not a json file. That way also analyzer would be covered, using different flags. But filtering out can frames based on the json test file is not in place (yet).

[testais.zip](https://github.com/canboat/canboat/files/3064485/testais.zip)

larsujensen:
I have included a python script, aisverify, for regression test of the ais part of n2kd in the zip archive which I upload again. The test is done by:
cat json_reg_test_file | n2kd -d 2>&1 | ./aisverify json_reg_test_file_output
It will output "true" if the test is successful, and "false" otherwise with some test data. A full description is given in a readme document in the archive.
The document also describes another way of using the script which may partly obsolete the need to generate a can file from the json_reg_test_file. It also describes the usage of the previous aisfilter_unique.
The archive now also contains the two candump files producing the test data.

And Teppo, I think this is about as close as I will get to the test data and test you asked for.

[testais.zip](https://github.com/canboat/canboat/files/3075315/testais.zip)

larsujensen:
I have filed an issue #161 regarding missing field "AIS Transceiver information" in PGNs 129809 and 129810. AIS Transceiver information gives the source of the message, own or other vessel. These common pgns are static data reports for class B ais transponders, providing information such as e.g. ship name, type of ship, and call sign. Rather than skipping these PGNs, the code defaults to assume it is from other vessel, but it might give a problem in some plotters. (It is possible to deduce origin based on MMSI, and I think it is possible to set this in configuration in opencpn, but this may not be the general case.) It should be easy to add AIS Transceiver information in the pgns, thus avoiding sending out false NMEA 0183 ais sentences.

PR #173, opened 2019-05-21
fix make file for ubuntu/debian packaging

sailoog:
I want to create and maintain a debian/ubuntu repository for some packages used in OpenPlotter:

https://launchpad.net/~sailoog/+archive/ubuntu/openplotter

Canboat is there but I have found some problems that make impossible creating a debia/ubuntu package to distribute it. This PR fix the problem.

- canboat install programs in usr/local/bin and this is forbidden in debian packaging. You can force it and create a deb but you can not force this to create a source for distribution. This fix uses /usr/bin
- canboat creates files in /usr/local/etc/default and that folder does not exist there. This fix uses /etc/default

Feel free to add this repo to your installation manual as an extra option if you want. Tested in Ubuntu 16.04 and Raspbian.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,10 +8,10 @@
 
 # s. https://www.gnu.org/prep/standards/html_node/Directory-Variables.html#Directory-Variables
 DESTDIR ?= ""
-PREFIX ?= /usr/local
+PREFIX ?= /usr
 EXEC_PREFIX ?= $(PREFIX)
 BINDIR=$(EXEC_PREFIX)/bin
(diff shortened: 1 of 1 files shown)


keesverruijt:
I will merge now but don't have time to make a release containing this right now, will take 1-2 weeks.

sailoog:
OK no problem. Remember to change the wiki, I think /usr/local/bin is here https://github.com/canboat/canboat/wiki/Building and maybe somewhere else. Thanks!

Issue #181, opened 2019-09-07
broken parsing of 126720


mbj4668:
It seems e9ae88f breaks parsing of pgn 126720.  Before this commit, the attached packets were parsed as:

    08:30:53.868 7   1 255 126720 Manufacturer Proprietary fast-packet addressed:  Manufacturer Code = Garmin; Industry Code = Marine; Data = 10 17 04 04 0e 04 0a f6 03 11 01 19 01 21 01

After this commit it is parsed as:

    08:30:53.867 7   1 255 126720 Manufacturer Proprietary fast-packet addressed:  Manufacturer Code = Unknown Manufacturer 416; Industry Code = Global; Data = 0x404171098e5
    08:30:53.868 7   1 255 126720 Manufacturer Proprietary fast-packet addressed:  Manufacturer Code = Unknown Manufacturer 1697; Industry Code = Global; Data = 0x11103f60a04
    08:30:53.868 7   1 255 126720 Manufacturer Proprietary fast-packet addressed:  Manufacturer Code = Unknown Manufacturer 418; Industry Code = Global; Data = 0xffffff012101

The reason seems to be incorrect (?) "size" in the definition of the Fusion PGNs.  The function printPacket() in analyzer.c seems to use the size from the first pgn definition with the given pgn number.  In this case it is 3 and not treated as a fast packet.  If this size is changed to 9 it kind of works, but it doesn't seem to be the right solution.

[x.txt](https://github.com/canboat/canboat/files/3586457/x.txt)

PR #186, opened 2020-01-19
Raymarine Seatalk1 Smart Remote behind a Raymarine Seatalk1 -> SeatalkNG converter

htool:
'Wind' and 'Track' button and PGN to set Pilot Mode through Raymarine Seatalk1 <-> SeatalkNG converter.

The order of the pgn.h is important as I ran into https://github.com/canboat/canboat/issues/181 which broke PGN 126720 for me.


keesverruijt:
Why did you remove the Simnet Pilot command?

htool:
I messed something up today in my workflow.

htool:
This commit: dd584eed1b812a3235fe34236463bef02c8ddc3b should be good. It should only add things.

PR #189, opened 2020-03-03
feature: adds support for alert pgns 126983, 126984, 126985

jncarter123:


keesverruijt:
@jncarter123 thanks, looks very good. I will merge immediately.

Do you have a "sample" file you can share? What device do you have that generates these?

jncarter123:
No problem. i'm using Maretron N2KView to generate the alerts.  I'll add some more sample data here. https://gist.github.com/jncarter123/2b9d0b7030920babc88e0a35de8ec16c

PR #206, opened 2020-08-05
Fix missing comma in definition for 127513

fabdrol:
Jeremy Carter and I in the Signal K slack group were debugging some buggy behaviour with the parsing of PGN 127513 (Battery Status Configuration). We believe it to be a missing comma. This PR seems to fix the issue, although we've not been able to test beyond his case.


fabdrol:
@keesverruijt thanks for merging. Was the updated `pgns.json` also pushed to `npm`?

fabdrol:
I've had a chance to test this some more with a bunch of maretron DCM100's, this issues seems to have been fixed.

PR #211, opened 2020-08-25
fix: 130314 actual pressure is a signed value

jncarter123:
130314 actual pressure allows for the measurement of vacuum. This change
allows for the negative values to be correctly decoded.

I will post some raw data from a Maretron FPM100.


jncarter123:
https://gist.github.com/jncarter123/0b693cb0414bda51eeebe1baa03008a5

PR #216, opened 2020-12-28
correct lat/lon field type assignments for PGN 129041

cbatson:
The definition for PGN 129041 had the types of the latitude and longitude fields inverted.


analyzer/pgns.json
```diff
@@ -7910,14 +7910,6 @@
...
           {
             "Order":3,
-            "Id":"installationDescription3",
-            "Name":"Installation Description #3",
-            "BitLength":16,
-            "BitStart":0,
-            "Type":"ASCII or UNICODE string starting with length and control byte",
-            "Signed":false},
-          {
-            "Order":4,
```
cbatson, line 7920: It appears this was out of date.

PR #245, opened 2021-07-28
Updates for PGNs 126993, 129541, and companyList

arks-public:
Hello everyone! This PR mainly brings improvements to PGNs 126993 and 129541, as well as the list of manufacturers.

- For the heartbeat message (126993), the linked NMEA corrigendum shows that the resolution of the Data transmit offset has been lowered from 10 ms to 1 ms in NMEA versions 2.100 above. Additionally, the size of the reserved field has been adjusted, and fields have been added for CAN controller statuses and equipment status as reflected in Maretron's N2KAnalyzer V3.
- For Inverter Status (127509), additional field descriptors have been added to the PGN based on the provided corrigendum.
- For GPS Almanac Data (129541), the field sizes have been adjusted to match the fixed packet size (26 bytes), with field sizes deduced from N2KAnalyzer. Some of the resolutions may need to be adjusted, since some fields like Rate of Right Ascension and Clock Parameter 2 have units of 10^-38, which produces floating point strings with a lot of leading zeros. See https://www.gps.gov/technical/icwg/IS-GPS-200M.pdf page 116 for more details on the applicable units and precisions.
- A proprietary Maretron temperature message was added (130823) based on materials found online. Can't vouch for this one 100% but it is a decent starting point.
- Lastly, the list of companies was updated according to the most recent list of manufacturer codes from NMEA (June 23, 2019).


analyzer/pgn.h
```diff
@@ -3141,18 +3141,20 @@ Pgn pgnList[] = {
...
+    /* https://www.nmea.org/Assets/20140102%20nmea-2000-127509%20pgn%20corrigendum.pdf */
     ,
     {"Inverter Status",
      127509,
      PACKET_COMPLETE,
      PACKET_SINGLE,
      4,
      0,
-     {{"Instance", BYTES(1), 1, false, 0, ""},
+     {{"Inverter Instance", BYTES(1), 1, false, 0, ""},
```
patg, line 3153: Why 'Inverter Instance'. All other Messages translate the NMEA2000 instance into 'Instance'. 
I can't see why 127509 should be an exception here.  Would be nice to be changed back to
'Instance' (not only because it breaks my code ;-)   )

arks-public: It's probably because I didn't have live data for this PGN, so I tried to match the [source](https://www.nmea.org/Assets/20140102%20nmea-2000-127509%20pgn%20corrigendum.pdf) document as closely as possible. Maybe I thought an "Instance" field would be ambiguous since the PGN also contains "AC Instance" and "DC Instance". That being said, I agree that it should stay consistent with other PGNs, so if you submit a pull request for this you'll have my support.

keesverruijt: I agree with @patg.

keesverruijt: Fixed in master.

PR #257, opened 2021-09-28
DBC file exporter

mairas:
# Introduction

This PR adds a DBC exporting script to the canboat repo. The DBC file format is a proprietary (but industry standard) means of describing CAN bus messages, supported by most generic and automotive CAN analysis and reverse engineering tools.

Having a DBC exporter and the DBC database file available for canboat allows for decoding NMEA 2000 messages using generic CAN tools.

# Implementation

I decided to implement the DBC exporter as Python script due to better library availability. Proper Python packaging is provided for the script, but the Makefile integration only installs the script in a virtual environment within the build tree.

# Limitations

The DBC file format is geared towards simpler CAN protocols and can't support all NMEA 2000 features. In particular, fast packets are not supported, and neither are ISO 11783 multi-packets. Fast packets _are_ detected, though, so at least the user will see the PGN being transmitted. PGN reuse for different messages is partially supported but quite hacky.


keesverruijt:
Super. I don't care for python myself but don't mind accepting this and merging this.

PR #258, opened 2021-09-30
B&G H5000 CPU performance packet

htool:
Packets sent from B&G H5000 CPU with performance info.
The packet seems to contain key-value pairs, hence the empty fields.
Wind Phase and Wind Lift are also encoded here, but I don't understand yet how.


htool:
Excuse the git mess - I'm still not comfortable with it :/

htool:
I added the other values by confirming changes to them are seen on the B&G MFD (Vulcan).
Wind Lift and Wind Phase look like radian values, but are offset against a value before displaying. Not sure yet how to get the exact value, but I am sure it's this key/value pair.

keesverruijt:
Great work,  but please check the PACKET_INCOMPLETE and PACKET_FAST flags,
not all look correct.

Op vr 1 okt. 2021 10:03 schreef Hans ***@***.***>:

htool:
Can you point me to where I read what they mean? So far I didn't find what was expected here.
I thought:
PACKET_FAST: multiple packets (as other option seems PACKET_SINGLE)
PACKET_INCOMPLETE: multiple options possible for this PGN.

htool:
Some sample data here: https://pastebin.com/ggyFjyMW

htool:
It looks like the key value pairs can change place, so this won't work.

keesverruijt:
Hi Hans, sorry for not reacting earlier but I didn't have a laptop with me. You are correct on the FAST/SINGLE flag, but the Incomplete flag means that "we" (the reverse engineers) feel that the packet analysis is not yet complete.

I will eagerly await your next PR!

PR #260, opened 2021-10-05
Update manufacturer list

rkuris:
I scraped the lastest manufacturer list and ran it through
a perl script to produce an updated list.

https://www.nmea.org/Assets/20190623%20manu%20code%20and%20product%20code.pdf
copy/paste then
perl -ne 'if (/(.*?) ([0-9A-F]+) (\d+) /) {print qq(\t\t{"$1", $3},\n)}'


keesverruijt:
Interestingly there is a spelling error in the source document, 'Chetco Digitial' which now seeps through.

Issue #266, opened 2021-12-28
Refactor data definitions


keesverruijt:
Refactor the data definitions for the PGNs so that:

* The C code becomes easier to maintain and modify.
* The XML data contains structured data so that it is easier to generate other specification languages from it. This then becomes useful for #265 .
* The existing XML and JSON code is not impacted other than new keys are added.

In particular, all regularity that there is to the datatypes possible in PGN fields should be exposed and exploited.

This will include a split of the current `analyzer` executable into an `analyzer` and `analyzer-generate-xml` program.

To accomplish this we will use two 'new to this repository' features of C99:

* C99 named field assignments: ` { .length = BYTES(2) ... }`
* Special smaller `.h` files that only contain macro calls. Such files can be called repeatedly with different definitions for the macro.

aldas:
Have you considered in addition to have documentation page like GPSD has https://gpsd.gitlab.io/gpsd/NMEA.html ? 
I'm asking this because GPSd page has useful notes along with sentences that help you understand the context. And it seems to be somewhat goto place where to look NMEA0183 related info.

ala some markdown file in repo or in wiki part that contains more human readable form of PGNs / types / resolutions. Although there are atm ~343 PGNs (247 unique) which would be monumental task to write down initially all by hand (unless generated to readable blocks or maybe Doxygen can do something) ... and probably hard to be disciplined enough to keep up to date.

I could help if needed.

keesverruijt:
Well, `analyzer -explain` goes a long way. Making this generate markdown, which is readable by humans, is a good plan.

You could help out once #266 is fixed by enhancing the `analyzer-explain -explain` code.

TwoClocks:
This is much easier to work with. thanks!
I will give you more feedback as I work with it. Do you want problems I find here in this thread or open issues?

Is the intent to be backward compatible? That's why enums show up both in the enumerations section and in individual PGNs?

TwoClocks:
A question. From the XML how would I tell the difference between
`field->resolution == RES_STRINGLZ` and `field->resolution == RES_STRINGLAU` and `field->resolution == RES_STRING` They need to be parsed differently, I think.

Ideally, I'd just like the raw value of `field->resolution`. Or the `#define` name. Hard to do the latter w/o reflection though. And maybe `field->units` as well. I think I need to parse them to generate the KSY.

aldas:
[...]

Issue #265, opened 2021-12-18
Proposal : Move pgn definitions from .h file to Kaitai Struct


TwoClocks:
First off, I'd like to say that I've been using this project and its data for years. I am very appreciative of the time and energy that has gone into creating and maintaining it.

For those unfamiliar, [Kaitai Struct](https://kaitai.io/) is a  declarative language for parsing binary data. It can output parsing code in 15-ish different programming languages. It's used fairly extensively for network protocols and file formats. You can see a list of published parsers here : [https://formats.kaitai.io/](https://formats.kaitai.io/)

I have taken a quick stab at auto-converting CANBoat's JSON data to a kaitai struct's `.ksy` file format. Here is one pgn in the format: [gnss_position_data](https://gist.github.com/TwoClocks/d6c443d7a33b502447d249b3e56cfa3b).
I generated this programmatically. I'm sure my JSON converted isn't complete and doesn't cover all the types correctly. I just wanted to do this as an experiment.

I also created (by-hand) a .ksy file describing yacht devices' [Voyage Recorder file format](https://gist.github.com/TwoClocks/dfacac137a9ac72d9f7523985d729935). 

With these two files and the kaitai struct compiler, I could parse a yatchd data file, and pull out all the position messages. I did it in both java and python w/ no changes to the `.ksy` files, and identical output. Each version ended up being a few lines of code (not counting all the auto-generated code). Here is a kotlin app that runs [though a yachd file](https://gist.github.com/TwoClocks/9c0892480005fb9e18df0813f829fb41).

As a bonus, it can also output graphviz docs for a `.ksy` file : [gnss-image](https://user-images.githubusercontent.com/5883156/146653681-398cca6d-2cf5-4bd6-a15d-2323a654a5f7.png)

Kaitai Struct also has a great [on-line IDE](https://ide.kaitai.io/) that's very helpful for reverse engineering data. That is what I used to create the voyage recorder parser. I was inspired by this article on using the IDE to reverse engineer : [https://gettocode.com/2017/09/18/reverse-engineering-with-kaitai-struct/](https://gettocode.com/2017/09/18/reverse-engineering-with-kaitai-struct/)

So, with all that being said : 

I propose that we permanently move the `pgn.h` to .ksy files, and retire `pgn.h`. It should be easy to port `analyzer` (and the rest of the tools in this project) to use generated C files from the kaitai struct compiler instead of `pgn.h`

[...]

PR #275, opened 2022-07-26
Refactor definitions fixes

Selebrator:
# Introducing myself
Hi, I'm Lukas, a Student at the University of Oldenburg, Germany.
I'm working on a project where we need to read and write NMEA 2000 messages. For that we want to use CANboat's definitions to generate Java code. Reading #265 and #266 I decided to go with the version from the refactor-definitions branch, as it seems great for code generation.

# What this PR does
- The generated json files contained a syntax error. That error is fixed. It's my first time seeing xslt, so the solution my not be ideal.
-  PGN 128776 Field 13 was incorrectly declared as a Lookup Table. In master it is a Bitfield. I reverted it to how it is in master.

My Team and I consider making more contributions to CANboat.

### DIFF
--- a/analyzer/pgn.h
+++ b/analyzer/pgn.h
@@ -2710,7 +2710,7 @@ Pgn pgnList[] = {
        false,
        "s",
        "If timeout elapses the thruster stops operating and reverts to static mode"},
-      LOOKUP_FIELD("Windlass Control Events", 4, WINDLASS_CONTROL),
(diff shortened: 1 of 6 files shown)


keesverruijt:
Thanks Lukas,

You have re-ignited this branch. I've been sailing so interest was elsewhere. Lets see if we can get this into production quality.

One thing I am considering now that I have re-read everything is that the `explain` program could write out 2 different XML files, one backwards compatible and one new format.

PR #356, opened 2022-12-05
Add position fixing device to AIS msg 24 B

Selebrator:
I'm not sure about the order, but this order best fits my examples. Looking at the [official AIS documentation](https://www.itu.int/dms_pubrec/itu-r/rec/m/R-REC-M.1371-5-201402-I!!PDF-E.pdf) I would have assumed them to be the other way around.

In my examples I observe, that the relevant byte is always one of `03`, `13` or `f3`. My assumption is that the lower octet wich is always `3` comes from reserved = empty and spare = empty.
Then the higher octet is `0` = undefined, `1` = GPS or `f` = internal GNSS.
```
2021-09-02T11:07:57.294Z,6,129810,4,255,35,18,18,84,98,0c,25,53,52,54,48,58,2a,34,44,43,37,33,35,36,40,50,00,1e,00,0a,00,1e,00,00,00,00,00,f3,e3,ff
2022-06-02T13:43:41.200Z,6,129810,5,255,35,18,0a,45,9a,0c,35,4e,41,56,49,43,4f,40,44,46,36,36,32,33,40,4a,01,3c,00,1e,00,04,01,00,00,00,00,03,e1,ff
2022-06-15T08:33:52.223Z,6,129810,5,255,35,18,e6,7f,9c,0c,25,57,44,40,44,40,5a,4e,44,47,33,39,31,33,40,96,00,28,00,0a,00,50,00,00,00,00,00,13,e0,ff
```


keesverruijt (commented):
I will try to verify asap. It looks plausible!

TwoCanPlugIn:
My observations from two different AIS transponders (Icom, Digital Yacht) are the following:

gpsFixingDevice = byte[33] & 0x1F; 
reserved = (byte[33] & 0xE0) >> 5; 
sequenceIdentifier = byte[34];

Apologies that this isn't in canboat format, but I'm sure it makes sense to you.

For the type of fixing device I've only seen values of 0 (which corresponds to the ITU "undefined/default" and 1 which corresponds to "GPS". 

I guess that by using 5 bits it accomodates the ITU definitions from 0 "undefined/defaut" up to 15 "Internal GNSS", and leaves extra bits to define the usual NMEA 2000 "data unavailable" values etc. 

Looking at the above samples, I would suggest that the first message, the fixing device (3) corresponds to combined "GPS/GLONASS", the second (1) corresponds to "GPS" and the third (0) corresponds to "undefined/default". It seems that the AIS receiver that generated these messages does not generate or use meaningful sequence numbers. 

If you need some additional samples (which are in candump format), just let me know.

keesverruijt:
Sorry, I just didn't have time to look at this yet other than recognising that this is coming from updated ITU 1371 standards; in 1371-5 this has been added. I will try to determine exactly what the bit ordering is in the updated N2K messages by using a completely up-to-date N2K device...

TwoCanPlugIn:
The standard is relatively old, dates from 2014 !

Here are two candump samples for you, both containing examples of PGN 129810, Class B, Static Data Part B. melbourne.log uses a Digital Yacht AIS transponder and seattle.log (which I've had to zip as too large to upload) uses an Icom AIS transceiver. 

Note that the seattle log also contains erroneous  PGN 129041 messages as I discovered that Icom incorrectly encoded the AToN Name field. After notifying Icom of the bug they subsequently released a firmware upgrade.

Hope this helps.
[melbourne.log](https://github.com/canboat/canboat/files/10638662/melbourne.log)
[seattle.zip](https://github.com/canboat/canboat/files/10638787/seattle.zip)

Issue #325, opened 2022-10-18
Automatic changelog creation


tkurki:
I've been pretty happy with PR-based changelog creation. In many Signal K repos it works so that
Prerequisites:
- Github repo's issue/pr labels are modified to include the values one wants to use
- (needed GH actions are added to repo)

Then:
- PRs are labeled (feature, fix, chore, other, ignore - how you want the changelog to be grouped)
- GH action checks that PRs have the label
- maintainer creates a GH release (can be automated off a pushed git tag with GH action also)
- GH action creates the changelog from PRs merged since the previous release

The last two release of [sk server] (https://github.com/SignalK/signalk-server/releases) use this method, earlier ones different tooling but essentially the same logic: changelog is created from PRs.

I can add the actions if this is something @keesverruijt you'd be interested in.

keesverruijt:
It would need careful implementation and operation, there are three things interacting here:

1. Generated XML and JSON files, especially when there are interacting PRs.
2. Creating a release.
3. Changelog.

If we go for GH actions, I would prefer it if the workflow refused straight changes to master other than CI changes. CI changes would be a pain to do via PRs as sometimes I need a lot of attempts at "will this work then to get the builder working?")

Also would be good if the version number of master would always be "unreleased"/"wip" and the versions live on a branch -- either "released" or a branch per version, although I don't see work on multiple branches being needed here.

If you are willing to do this for me, and explain in detail how this then works, great.

keesverruijt:
I guess I just proved why this is necessary, I accidentally referred to the wrong issue in a commit message.

tkurki:
The definite version information is in version.h, right? Is everything else derived stuff from that? So in master it could be `unreleased`. To create a release you'd decide the next version, update version.h and commit that locally, create a tag and push it. (or even that can be automated, the release GH action would update version.h off the tag).

Re XML and JSON files: I would let the dog wag the tail, have the release process generate them. So they would not be in git, but available as part of the release artifacts. (the default artifact is just the repo contents in a zip file, but can include other files also).

There's also the option to have not only a single, latest version at https://canboat.github.io/canboat/canboat.xml, but like SK spec: https://signalk.org/specification/latest/ redirects to the latest version 1.7.0, but all the published versions are still available and addressable, for example https://signalk.org/specification/1.5.0/doc/. The release process could automate all this.

aldas:
> Re XML and JSON files: I would let the dog wag the tail, have the release process generate them

[...]

PR #369, opened 2023-02-02
Create Release on new tag and add PGN databases as artifacts

aldas:
This is related to https://github.com/canboat/canboat/issues/325

This Github workflow is triggered when new Tag with pattern `v[0-9]+.[0-9]+.[0-9]+` is pushed/created.   As a result https://github.com/softprops/action-gh-release actions will create new Release with that tag name and attach following files as artifacts `docs/canboat.json`, `docs/canboat.xml`, `docs/canboat.xsd`.

Release will look like that:
image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,12 +9,16 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: checkout repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
+
       - name: update & install tools
         run: sudo apt-get update && sudo apt-get install -y xsltproc libxml2-utils
+
       - name: make
         run: make
+
       - name: make generated
         run: make generated
+
       - name: check there is no git diff output
         run: git diff --exit-code
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,24 @@
+name: Upload release assets
+
+# this flow will be run only when new tags are pushed that match our pattern
+on:
+  push:
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Add release assets
+        uses: softprops/action-gh-release@v1
+        with:
+          files: |
+            docs/canboat.json
+            docs/canboat.xml
+            docs/canboat.xsd
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -15,6 +15,20 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
+      - name: Version in common/version.h must match release tag
+        run: |
+          export CANBOAT_VERSION=$(sed -En 's/.*\ VERSION\ \"([0-9]+\.)([0-9]+\.)?([0-9]+)\"/\1\2\3/p' common/version.h)
+          echo "Checking if docs/canboat.json contains \"CANboat version v${CANBOAT_VERSION}\""
+          grep -q "CANboat version v${CANBOAT_VERSION}" docs/canboat.json
+          echo "Checking if docs/canboat.json contains version \"Version\":\"${CANBOAT_VERSION}\""
+          grep -q "\"Version\":\"${CANBOAT_VERSION}\"" docs/canboat.json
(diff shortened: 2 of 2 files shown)


tkurki:
Nice! 

The update to checkout v3 in ci.yml could be a separate commit, as it is not really a part of this feature.

I think it would be a good idea to sanity check that the versions in .json and .xml files matches the tag to guard against human errors.

aldas:
I think sanity checks should be done for PRs or commits to master. Blocking release page being automatically being updated does not solve the (root) problem - which is commits or PR were untested.

This flow starts when tag is created/pushed. I am not aware if Github workflow supports "pre-tag-hooks" that block creating/pushing tags. Because if this workflow would stop run on problems, this would mean only "release" page is guarded - the problematic Tag still exists and problematic commits still exist in git log/history.

tkurki:
If you by mistake create a tag that does not match the version in the files then you also create an inconsistent release. A check in this workflow would alert you to the mistake and prevent creating the release with inconsistent information, allowing you to take fix things by fixing just one mistake (remove the bad tag) and not two (bad tag and bad release).

You can't check if the version in the files is the same as the tag in master nor in PRs, because neither of them are the tag. A pre push hook that runs on tags would probably work.

aldas:
These are good things but are not necessary for this PR as it is not meant to fix quality assurance problems or complicate current developer way of working. 

This is a small incremental step in forward - it solves one specific thing without affecting anything else. If additional/different quality gates or changes in development process are needed these could be added in further PRs as involved people see it fit.

keesverruijt:
I like this (small but significant step)!

> I am not aware if Github workflow supports "pre-tag-hooks" that block creating/pushing tags. 
I don't think so, as @tkurki says a `git commit` with a version number in a file is independent of pushing the tags.

> Because if this workflow would stop run on problems, this would mean only "release" page is guarded - the problematic Tag still exists and problematic commits still exist in git log/history.

I agree with @tkurki : anything automated should check for all it can, and this would prevent silly mistakes -- even if it is only the release page. I still think it is worthwhile.

A good next step could be: only push to [github.io](https://canboat.github.io/canboat/canboat.xml) when a release is made, so that intermediate commits aren't pushed all the time.

aldas:
At the current development process maturity level, schema version is probably one of the smallest things to care in relation to schema files. 

There have been changes to schema - to struct fields or classifier values etc so you always need to check in your own library/application if can "support" that schema at all. So you pretty much are forced to lock your library/application to one fixed version that you have tested. In that case whatever number is in xml/json does not matter at all. As currently no semantic versioning actually exists or is being practiced - means that major/minor/patch parts in schema version are useless. 

Some example: recently `STRING_VAR` field type was removed, `INDIRECT_LOOKUP` type was added - any change to field type is more than likely breaking change and must trigger major version bump. NB: Schema version and Canboat version is different.

Unless strict semantic versioning is being done and CI flow to detect these changes are in place - checking version string when tag is pushed makes little difference.  

Semantic versioning is hard to achieve as it very hard to check with static analysis - it is more to do with developer own discipline and maybe review process. 

Allowing downstream users to pin to fixed version - gives them some guarantees that only verified/test (by themselves) version will be used/downloaded.

tkurki:
Did not want to start a debate, merely tried to suggest something that I thought would be a good idea and very little more work.

tkurki:
Ok, one more comment, sorry 😄 

Afai understand if this PR is merged one is supposed to git tag versions and once pushed that will create automatically Github Releases containing the artifacts. I think that is a great idea.

My concern is that if there is no sanity check that ensures that the artifacts have the same version number as the release merging this PR creates a mechanism where it is pretty easy to create inconsistent releases. This is directly related to the mechanism introduced in this PR, not a separate issue.

Even today the situation is that if I download release artifacts for [release v1.3.0](https://github.com/canboat/canboat/releases/tag/v1.3.0) I get files where it says `#define VERSION "1.2.1"`.  Errors like this are easy to do when there are manual steps. Been there, done that, several times.

How versioning happens is a separate issue, but ensuring that tags & artifact contents match is valuable in its own right.


But at the end of the day I am still 👍 for merging this and starting using GH Releases, check or no check.

aldas:
I have added some checks to compare `common/version.h` value with artifacts.

I do no think this is eloquent solution because only thing this does is that Release page will not be updated. The "error flow" is limited in this situation.

The TAG still exists so that buggy release *actually* exists.  Github releases are just cosmetic Github specific feature that is built on Git tags and help you download repository artifacts from web (instead of cloning the repo etc).

PR #376, opened 2023-02-10
PGN 130571 fix

mbj4668:
This PR adds field 3, "Type" to PGN 130571.   See the file "https://www.nmea.org/Assets/20160715%20corrigenda%20entertainment%20pgns%20.pdf" (except that the link doesn't work anymore...)

I haven't seen this PGN, but found the error in my code, which uses canboat.xml to generate code.


aldas:
Some of the NMEA hosted documents can be found at Wayback Machine 

https://web.archive.org/web/20170609110901/https://www.nmea.org/Assets/20160715%20corrigenda%20entertainment%20pgns%20.pdf

mbj4668:
Thanks!  Some of them can also be found as HTML in the google search cache.

PR #378, opened 2023-02-13
Change PGN 60928 (ISO Address Claim) last field to "Arbitrary address capable"

aldas:
# Change PGN 60928 (ISO Address Claim) last field to "Arbitrary address capable"

`Arbitrary address capable` field indicates whether the device is capable to claim arbitrary source  address. Value is 1 for NMEA200 devices. Could be 0 for J1939 device claims.  See: https://embeddedflakes.com/network-management-in-sae-j1939/#Arbitrary_Address_Capable

This is useful in case you are debugging/analyzing bus for the first time and this field helps better to understand what kind of device(s) you are dealing with.


aldas:
ok, done, moved docker stuff to another PR.

PR #384, opened 2023-03-07
Updated lengths of lighting PGNs.

mglonnro:


mglonnro:
For `130330` I added `RESERVED_FIELD(2)` to the end to fill out the total number of bits to divisible / 8. Don't know if that was the correct thing to do 🤷‍♂️

mglonnro:
> > For `130330` I added `RESERVED_FIELD(2)` to the end to fill out the total number of bits to divisible / 8. Don't know if that was the correct thing to do 🤷‍♂️
> 
> Normally it would but I think you miscounted. You are changing the fields in the first byte but the rest stays the same so I am not seeing an non-byte-sized number of bits. Note that various fields such as strings must start on a byte boundary.

Ah, you are right! 

> Can you find out what the various values of field 2 of PGN 130330, `Default Settings/Commands` are? This would then be encoded as
> 
> ```
>  LOOKUP_FIELD("Default Settings/Command", 3, LIGHTING_COMMAND)
> ```
> 
> in `pgn.h` and the lookup `LIGHTING_COMMAND` needs to be added to `lookup.h`.
> 
> Even if you have only 1 or 2 values that you know this is already useful.

Will do!

PR #390, opened 2023-03-25
Add missing reserved field to 130312

sbender9:


keesverruijt:
It's okay if pull requests don't update the generated files; if you don't I will. If you do, please update CHANGELOG.md as well.

PR #420, opened 2023-09-08
fix: add fusion Album message id lookup

sbender9:


sbender9:
I think the test failure is not related my change, right?

keesverruijt:
> I think the test failure is not related my change, right?

Nope, the dbc-exporter did not fix the cantons version so a refactoring of cantools broke the dbc-exporter. This is now fixed.

keesverruijt:
I'll merge but FYI best practices is to run `make generated` and add to CHANGELOG.md as well.

PR #424, opened 2023-09-28
Add Field_Index, complete FIELDTYPE_LOOKUP

inquestruss:
I added a new FieldType: "FIELD_INDEX", which alerts our code generator to pass the field's value, along with the reference PGN and ManufacturerCode, to the routine that decodes the following "VARIABLE" field. I've also changed the FieldTypes for the keys of KEY_VALUEs to FIELDTYPE_LOOKUP, for the same reason.


analyzer/pgn.h
```diff
@@ -130,7 +130,7 @@ typedef struct
 #define LOOKUP_FIELDTYPE_FIELD(nam, len, typ)                                                          \
   {                                                                                                    \
     .name = nam, .size = len, .resolution = 1, .hasSign = false, .lookup.type = LOOKUP_TYPE_FIELDTYPE, \
-    LOOKUP_FIELDTYPE_MEMBER = lookup##typ, .lookup.name = xstr(typ), .fieldType = "LOOKUP"             \
+    LOOKUP_FIELDTYPE_MEMBER = lookup##typ, .lookup.name = xstr(typ), .fieldType = "FIELDTYPE_LOOKUP"             \
```
meaning-matters, line 133: Should be `FIELD_TYPE_LOOKUP` because `fieldType` is two words.

analyzer/pgn.h
```diff
@@ -488,6 +494,11 @@ typedef struct
     .name = nam, .size = BYTES(1), .resolution = 1, .fieldType = "UINT8", .description = desc \
   }
 
+#define FIELD_INDEX(nam, desc)                                                           \
```
meaning-matters, line 497: Why `FIELD_INDEX` and not `FIELD_TYPE_INDEX` similar to `LOOKUP`?

inquestruss:
Any names are fine with me--I didn't think about them carefully, sorry.
FIELD_INDEX isn't about type, directly. It's the value of Order for the
target field, a 1-based index into the list of fields. Maybe FIELD_ORDER
would be better?

PR #430, opened 2023-12-11
Add select J1939 PGN to canboat/analyzer.

g2griz:
Certain Yamaha jetboats have a dedicated can bus for the engine ECU using J1939.  (Separate from the N2k can bus.) This pull will allow canboat/analyzer to process select J1939 messages for eventual display using on OpenPlotter via signalk.

An alternative to adding these PGNs to canboat/analyzer would be to create a distinct variant to only handle J1939 messages.  The code would be identical albeit with a much smaller pgh.h.


keesverruijt:
I'm a bit worried that there might be J1939 messages that have overlapping PGNs, and/or are in the proprietary NMEA2000 PGN ranges but that are not proprietary in J1939 -- I don't think J1939 has this concept? 

All of this means that a separate J1939 database would make more sense, in my opinion.

g2griz:
Thanks for the comment.  J1939 defines the same PGN ranges between proprietary & non-proprietary as NMEA2000.  They also share the same PGNs and protocols for ISO messages. Unfortunately without accesss to the NMEA & SAE standards I cannot verify if there are any conflicting PGN definitions.

I will build a J1939 version of canboat/analyzer using a separate pgn.h and see how that goes.

g2griz:
OK.  I created a separate pgn-j1939.h and added conditional include logic in analyzer.h based on -D flag.  Also updated Makefile to create new analyzer-j1939 target.

I'm not sure how to roll back to the original pgn.h as part of this pull request however...

keesverruijt:
> I'm not sure how to roll back to the original pgn.h as part of this pull request however...

Create an extra commit containing the rollback to the original pgn.h file, then use 

     git rebase -I HEAD~n

Where n is the number of commits in your branch, run the command. In the editor that shows reorder the last commit to just after the one where you edited pgn.h, and change it to “fixup”.

If you want to drop the first commit altogether you can just remove it in the editor started by the interactive rebase.

g2griz:
Please pardon my feeble command of git.  However, I think everything is in order now.

g2griz:
I corrected an error on the scaling for the temerture-related PGNs.

This commit should have the following six files:
analyzer/Makefile
analyzer/pgn.h
analyzer/fieldtype.h
analyzer/analyzer.h
analyzer/pgn-j1939.h

Many thanks for your help.  Let me know if you need anything else.

g2griz:
Is there any thing else I need to do on this?  Thanks

g2griz:
Thanks.  Have a great holilday

Issue #436, opened 2024-02-12
Argument for change to PGN 65379


cas206:
### Discussed in https://github.com/canboat/canboat/discussions/434

<div type='discussions-op-text'>

<sup>Originally posted by **cas206** December 22, 2023</sup>
From collected data: 

```
// Command (0 - 204)
2023-12-21T18:30:55.187Z,3,126208,0,204,17,
01, // Command
   63,ff,00,// PGN 0x00ff63 = 65379
   f8, // Priority 8
       // Reserved (f)
   04, // Number of Parameters 4
   01,3b,07,
   03,04,
   04,40,00,
   05,ff,ff

// Group Command Acknowledge
2023-12-21T18:30:55.187Z,3,126208,204,0,8,
   02, // Acknowledge
   63,ff,00, // PGN 0x00ff63 = 65379
   00, // Error Code
       // Transmission Interval
   04, // Number of Parameters
   00, // No Error 1
       // No Error 2
   00  // No Error 3
       // No Error 4

```
The PGN 126208 Command to PGN 65379 is treating the two Pilot Mode (8 bit) / Pilot Submode (8 bit) as a single 16 bit field.  This is followed by another 16 bit field.  Could the Pilot Mode be 16 bits and the Pilot Submode be 16 bits?  Or are they combined into a single 16 bit value, followed by 16 bit data?

The next PGN 65379 Data is: 

```
// PGN 65379
2023-12-21T18:30:55.187Z,7,65379,204,255,8,
   3b,9f, // Field 1,2,3
   40,00, // Field 4 (as seen in Command message)
   00,00, // Field 5 (as seen in Command message)
   02,    // Field 6 ?
   ff     // Field 7 ? / Reserved
```
</div>

PR #437, opened 2024-02-12
 Argument for change to PGN 65379

cas206:
Fixes canboat/canboat#436

Changed pgn.h for PGN 65379.  Both Modes are 2 bytes.  Reserved is 1 byte.


cas206:
Assuming this change is accepted, do you prefer only the source change and to run the generated files yourself?  Or should I add the changed generated files as well?

cas206:
Updated with generated files so that the automated check passes.

PR #154, opened 2019-02-27
PGN 65340, 65379 documentation

Tehani1:
I have found information about these PGNs. It's not the complete, but I think it's enough.

#define LOOKUP_SIMRAD_PILOT_MODE    \
  (",0=Standby"       \
   ",1=Auto, compass commanded"                  \
   ",2=Non Follow, hand commanded (not sure)"    \
   ",3=Wind Mode"                                \
   ",4=Track Mode"                               \
   ",5=No Drift, COG referenced")

#define LOOKUP_RAYMARINE_PILOT_MODE    \
  (",0x0000=Standby"       \
   ",0x0040=Auto, compass commanded"                  \
   ",0x0100=Vane, Wind Mode"                          \
   ",0x0180=Track Mode"                               \
   ",0x0181=No Drift, COG referenced (In track, course changes) ")

SECTION:
/* proprietary PDU2 (non addressed) single-frame range 0xFF00 to 0xFFFF (65280 - 65535) */
--------------------------------------------------------------------------------
	{"Simnet: Autopilot Mode",
     65340,              /* NOTE 65341 in pgn.h maybe is not correct */
     false,
     0x08,
     0,
     {{"Manufacturer Code", 11, RES_MANUFACTURER, false, "=1857", "Simrad"},
      {"Reserved", 2, RES_NOTUSED, false, 0, ""},
      {"Industry Code", 3, RES_LOOKUP, false, "=4", "Marine Industry"},
	  {"Autopilot Status 1", BYTES(1), false, "=0x10", "Pilot ON"},
	  {"Autopilot Mode", BYTES(1), false, LOOKUP_SIMRAD_PILOT_MODE, ""},
	  {0xfe},	  
      {"Autopilot Status 2", 3, false, "=2", "Pilot ON"},
	  {"Reserved", 5, RES_NOTUSED, false, 0, "0x1F"},
	  {"Reserved", BYTES(2), RES_BINARY, false, 0, "0x00, 0x80"}}}

// Capture StandBy: 0x0CFF3C00, 0x41 0x9F 0x00 0x00 0xFE 0xF8 0x00 0x80
// Capture Auto:    0x0CFF3C00, 0x41 0x9F 0x10 0x01 0xFE 0xFA 0x00 0x80
------------------------------------------------------------------------------------
	,
	{"Seatalk: Pilot Mode",
     65379,
     false,
     0x08,
     0,
     {{"Manufacturer Code", 11, RES_MANUFACTURER, false, "=1851", "Raymarine"},
[...]

PR #439, opened 2024-02-17
Updates related to Raymarine PGN 65379 + 1 minor fix to Makefile

cas206:
1) Added Lookup Table Values for Raymarine PGN 65379.
Refer to https://github.com/canboat/canboat/pull/154 for source of table.  I can confirm that "Standby" and "Auto, compass commanded" match with my test with EV-1 Autopilot.
 
2) Added raw files to samples folder for PGN 65379.
I'm not sure what raw files are for or the best way to present them.  Here is my initial attempt.  
 - A file with just PGN 65379 messages seems to be of limited use.  So I have a pgn65379.raw file with PGN 65379 and PGN 126208 which sends data for PGN 65379.  I cut out duplicated data messages and only include two mode switches, Auto->Standby, Standby->Auto. 
 - raymarine-ev1.raw which contains all the data recorded in samples/ydnu-02-raymarine-ev1.dle
 
3) Added test of analyzer for PGN 65379.
Added a basic test for the new PGN 65379 structure.
(unrelated - test8 seems to have been left out of the .PHONY line in Makefile.  I added it)


sbender9:
@cas206 are you 100% sure about this change? 65379 was working fine before your change, although with different interpretations of mode and submode. 

@keesverruijt I do not think this is correct, and it is a breaking change for Signal K

cas206:
> @cas206 are you 100% sure about this change? 65379 was working fine before your change, although with different interpretations of mode and submode.
> 
> @keesverruijt I am not sure this is correct, and it is a breaking change for Signal K

Am I 100% sure?  Never.  It's based entirely on how Raymarine splits the message up when addressing 65379 via the 126208 command.  126208 sends Field 4 as 16 bits and Field 5 as 16 bits.  It's undetermined if the last two bytes are a single field  (Field 6), or two fields (Field 6, 7).  

Before the PGN format change, "analyzer" could not correctly parse 126208.  After the change, it gives good output.

PR #460, opened 2024-11-17
Update pgn.h

rszemeti:
PGN127508 was incorrectly using an unsigned int/precision 10mv for votlage, correct field is signed int, precision 10mv


keesverruijt:
Fixed in #464.

PR #481, opened 2025-01-21
Added manufacturer code for Revatek

grevelle:


keesverruijt:
You added this to the wrong file. Also, you are using the wrong file in the first place, you should be using `docs/canboat.json` or `docs/canboat.xml` instead of `analyzer/pgns.json` or `analyzer/pgns.xml`. The latter will be removed shortly as they are in an old, inconsistent and incomplete format.

grevelle:
Thanks Kees - will use those next time.

PR #487, opened 2025-03-07
Adding PartOfPrimaryKey to canboat

tomer-w:
Based on the discussion [here](https://github.com/canboat/canboat/discussions/486). This is my 1st attempt with adding PartOfPrimaryKey.
I'm quite sure I need to make more changes, but I need some help with what is expected.


keesverruijt:
You will need to add a corresponding boolean field in the `Field` structure in `analyzer/pgn.h` and then set that in all PGN fields, e.g. you'd see 

    .primary_key = true

for these fields in the PGN list. Add the field to `docs/canboat.xsd` as well. Then, modify the generation process of the XML in C and add it to the conversion from XML to HTML and JSON (in `docs/canboat.xsl` and `analyzer/canboat2json.xsl`).

tomer-w:
Few questions:

1. I assumed ```analyzer/pgn.h``` is generated. isnt it the case?
2. I did update the ```canboat.xsd``` and ```canboat.xsl``` so the json file and the html file already include the new field. You can take a look on the PR content.

Issue #492, opened 2025-03-20
Fallback PGNs are missing from canbaot.json


tomer-w:
[My](https://github.com/tomer-w/nmea2000) python NMEA 2000 decoder which is based on canboat failed some tests. Apparently, as my data source is canboat.json, it is missing all the Fallback PGNs.

From canboat2json.xslt:
```xslt
<xsl:template match="PGNs">
    "PGNs":[
      <xsl:for-each select="*">
        <xsl:choose>
          <xsl:when test="./Fallback = 'true'">
          </xsl:when>
          <xsl:otherwise>
            <xsl:apply-templates/>
            <xsl:if test="following-sibling::*">,</xsl:if>
          </xsl:otherwise>
        </xsl:choose>
      </xsl:for-each>
    ]
  }</xsl:template>

```
I find it strange that ```./Fallback``` is filtered out. It means that my source of truth is partial. What is the reason to filter it out? Can I submit a fix to remove this choose clause?

keesverruijt:
> I’m fine with removing this filter, but as a separate PR not mixed in with your other quest. 
> 

I’m happy to remove it myself as well, without a PR.

tomer-w:
PR submitted: https://github.com/canboat/canboat/pull/493

keesverruijt:
Merged.

PR #493, opened 2025-03-21
Remove Fallback filter from canboat.json

tomer-w:
Change canboat2json.xslt to not filter out PGNs which has Fallback attribute

### DIFF
--- a/analyzer/canboat2json.xslt
+++ b/analyzer/canboat2json.xslt
@@ -463,14 +463,8 @@
   <xsl:template match="PGNs">
     "PGNs":[
       <xsl:for-each select="*">
-        <xsl:choose>
-          <xsl:when test="./Fallback = 'true'">
-          </xsl:when>
-          <xsl:otherwise>
-            <xsl:apply-templates/>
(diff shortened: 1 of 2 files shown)


tomer-w:
As discussed here: #492

sbender9:
FYI @keesverruijt , this is a breaking change for canboatjs.

sbender9:
I can fix it in code, of course, but then @canboat/pgns won't be backwards compatible with 5.x.

Bump to 6.x?

keesverruijt:
I was thinking 5.2 but I can live with 6.0, there are a few more such changes.

keesverruijt:
I don’t see why you can’t just always filter for all fallback PGNs if you don’t want them, even when parsing older files?

sbender9:
the problem is when older versions of canboatjs use the new files.

But I just checked, 5.2 is fine.

sbender9:
The current version of canboatjs depends specifically on 5.1.x of @canboat/pgns. I was thinking it depended on 5.x.

So go ahead with 5.2 if you want.

Issue #498, opened 2025-04-06
STRING_LZ field lengths should be explained better


Elmue:
You write for STRING_LZ:
"A varying length string containing single byte codepoints encoded with a length byte and terminating zero.
The length of the string is determined by a starting length byte."

But then you define a fix BitLength for all STRING_LZ as 40 or 80 or 96 or 104 or 256 bits, 

This is contradicting.
Please make clear if STRING_LZ is either fix length or variable length. 
It cannot be both at the same time.

And if it is variable length why do you define a fix BitLength ?
You never do this for STRING_LAU.

keesverruijt:
For most of these fields we think we have seen practical limits on the _maximum_ length of the STRING_LZ field. I think that is what it says for all these PGNs: we don't really know the maximum length. A safe implementation will allow for any length of (up to 233 bytes if a fast PGN is involved). For instance, when I reverse engineered the SonicHub PGNs (now the Fusion and generic audio PGNs) it seemed there was a maximum track and album name length much shorter than 233. 

For STRING_LAU I have more hints that the length is indeed unspecified and that the maximum is bounded by the max PGN size.

So what I will do is add a note to STRING_LZ is that the field length is just an observation in the field, that may be wrong.

Elmue:
Very inconsistent these guys from Fusion!
They introduce a length byte and then they don't use it ?
So how do I handle these LZ strings ?

I see that in all messages this type of string is the last signal in the message.
So I simply read to the end of the message and skip the preceding length byte ?

How do the strings look in the real world?
Do they have a fix length and they fill up the remaining bytes with zeros ?
Which logfile do you recommend me to study the data ?

keesverruijt:
TBH these are only defined for SonicHub (last produced in 2014 I think) and Fusion specific. They have been made obsolescent by the generic audio PGNs.

Maybe somebody can create a logfile.

I will update the doc to point out that we don't know.

tomer-w:
@keesverruijt , yes, this is my understanding based on their code and the new sample file.

TwoCanPlugIn:
I think it is fair to say that Fusion are an anomaly. (Their use of a proprietary Fast Message, PGN 126720 to send commands with less than 8 bytes in length attests to that!)

Here are samples I collected when reverse engineering PGN 130820 transmitted from a Fusion Media Player. 

[...]

PR #515, opened 2025-04-22
Fusion PGN 130820 parsing

tomer-w:
This is almost full parsing of Fusion PGN 130820. It is based on the code from the Android Fusion App. For now, I updated all PGNs which were already present in canboat and the ones which were supplied by @sbender9 (Thanks!!). I added this sample file to the samples directory, I hope it is fine.

@keesverruijt, please note that I have more info about https://github.com/canboat/canboat/issues/498. The Fusion code regarding ```STRING_LZ``` is quite clear and in all places it has no length limit.

Example of SIRIUS_CONTENT_INFO
(Variable names are my names just that it will be easy to follow the generated code. Function names coming from the actual code)

```
        public final t1 decodeFromByteStream(byte[] bArr) {
            t1 t1Var = new t1();
            StatusOpEnum decodeFromByteStream = StatusOpEnum.decodeFromByteStream(bArr);
            ve.t.checkNotNullExpressionValue(decodeFromByteStream, "decodeFromByteStream(...)");
            t1Var.setType(decodeFromByteStream);
            t1Var.setSourceID(TCP_receiver.getUInt8(bArr, 2));
            t1Var.setChannel(TCP_receiver.getUInt16(bArr, 3));
            t1Var.setName(TCP_receiver.getString(bArr, 5));
            return t1Var;
        }
```
Implementation of TCP_receiver.getString
```
    public static String getString(byte[] bArr, int startingOffset) {
        StringBuilder sb2;
        byte b10;
        if (bArr == null || bArr.length - startingOffset < 1) {
            sb2 = null;
        } else {
            int len = bArr[startingOffset] & 255;
            sb2 = new StringBuilder();
            int i10 = startingOffset + 1;
            try {
                if (i10 < bArr.length) {
                    sb2 = new StringBuilder(new String(Arrays.copyOfRange(bArr, i10, i10 + len), StandardCharsets.UTF_8));
                }
            } catch (Exception unused) {
                sb2 = new StringBuilder();
                while (i10 < bArr.length && i10 <= len + startingOffset + 1 && (b10 = bArr[i10]) != 0) {
                    sb2.append((char) b10);
                    i10++;
                }
            }
        }
        return sb2 == null ? "" : sb2.toString();
    }
```

So, I think we can make the documentation clearer about that. I didn't want to do it in this checkin, but I'm open to your comments.


keesverruijt:
Great work here, btw...

tomer-w:
@keesverruijt BTW: I think it is better to [squash-and-merge](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#squash-and-merge-your-commits) instead of regular merge. I dont think you need in the main branch all the interim junk from my work. This is usually how it is done in other OSS projects. at least to the best of my knowledge.

sbender9:
@keesverruijt I'd like to some real world testing with this before you merge, will do today or tomorrow...

sbender9:
A quick scan of the code and I see that this is going to be a real nightmare for me. Definitely going to break much of my code that deals with fusion stereos, in the sk plugin, n2k-signalk and WilhelmSK.

sbender9:
@tomer-w the biggest issue for me is the changing of existing field names. Think about any code out there in the world that uses this stuff... now broken. Can you try to do this in a way that existing code will continue to work?

tomer-w:
@sbender9 , all the names I chose where the exact names used by Garmin in their code. My assumption was that canboat should try to be as official as possible. Other than that I don't have strong opinion one way or the other. I will be glad to hear @keesverruijt thoughts about this. If you guys will want to do some renames, I am not objecting.
In general I think each library should be as accurate as possible and that integration of new version can have it's own issues. @keesverruijt just changed time to duration which also broke my library, but I think this was good as be improved canboat and we should spend the time to align as needed.

tomer-w:
Is it really more than couple of hours of work?

sbender9:
Yeah, it's a tough one. My opinion on this case is that there is not much benefit in being aligned with Garmin. Definitely not enough benefit to make it worthwhile to cause a lot of broken code to cause hours of work testing, debugging and fixing code. (And yes, I am being a bit selfish here, but the truth is that this would cause a significant delay in the next version of canboat getting into canboatjs and signalk)

keesverruijt:
I hadn't looked in depth, but yes I am in favour of keeping existing field names as well -- as long as they are not totally wrong. 

In hindsight, we should have started with internal field names and external display names at a much earlier time. 

We can still do that, or have an optional "display name" that would contain the Garmin field name.

sbender9:
Just make all the field names that already existed have the original names. Obviously, any new fields those A,B,C fields can be changed.

keesverruijt:
I concur

PR #520, opened 2025-04-24
Revert Fusion field names to their original value

tomer-w:
As discussed, I renamed back the ones which were not problematic. Few I kept as is.
For:
Fusion: Media
Fusion: Track Position
Fusion: Tuner
I made the changes, but I think Garmin ones are better. With the rest I have less problems.
Will be glad to revert the changes in those specific ones. Let me know what you think in the PR.


sbender9:
Looks like you missed "Source Type" in "Fusion: Source"

Just started testing, so probably more to come...

sbender9:
Can you please also revert "Message ID" changes?

The first one I am seeing is "Power State", was "Power".

Of course, this breaks my code.

PR #524, opened 2025-04-30
Fix remaining compatibility issues with recent fusion changes

sbender9:
@tomer-w please review


analyzer/pgn.h
```diff
@@ -6540,7 +6540,7 @@ Pgn pgnList[] = {
      PACKET_FAST,
      {COMPANY(419),
       MATCH_LOOKUP_FIELD("Message ID", BYTES(2), 32813, FUSION_STATUS_MESSAGE_ID),
-      UINT8_PRIMARY_KEY_FIELD("Source"),
+      UINT8_PRIMARY_KEY_FIELD("Number"),
```
tomer-w, line 6543: I don't like field with the name "number" here. Too much meaningless in my mind.

tomer-w (approved):
Other than my single comment, the change is ok and keeps the original meaning. For the long term readability I would tweak downstream code and keep the current field name in that specific case. Anyhow I'm fine with whatever you decide.

When you are done I have more changes coming for the set PGN. Hopefully it will have no conflicts with your work as it is mostly new.

sbender9:
I'm not using the Set pgns, so that should be fine. 

I think @keesverruijt idea about having separate "field names" and "display names" is great. We can fix all this up if we do that. 

@keesverruijt you want me to take a swing at that? Let me create an issue...

sbender9:
https://github.com/canboat/canboat/issues/526

Issue #526, opened 2025-05-01
We should have separate "field names" and "display names"


sbender9:
Sometimes we later find that the existing field names are not correct or could be more descriptive, but changing them will break existing code. 

Creating a new item in the definitions for display names will allow us to make changes without breaking code. 

Currently we generate an item called "Id", what if we just allow that to be specified in pgn.h?

Then anytime we want to change a "Name", we would hard code the old "Id" item so that it does not change. 

Some work on the sk side, but the nice thing is that this change won't break existing code and won't require any changes to the existing definitions.

keesverruijt:
Great idea. Of course, in hindsight, we should have defined technical field names in the first place.

PR #527, opened 2025-05-01
WIP, POC allow the camel name to be specified

sbender9:
POC to see the best way to allow the camel name to be specified

See #526

Need to deal with -upper-camel properly


sbender9:
The macro definitions will start getting pretty disgusting, now with _DESC and _CAMEL versions.

wondering if it is better to bite the bullet and add the param to the existing macros and pass NULL unless it is needed?

keesverruijt:
Let me brood a little on this — there may be some preprocessor Fu we can use.

keesverruijt:
How about using a nested macro call:

     #define ID(camel, nam) nam, .camelName = camel
     ...
     BINARY_FIELD(ID("someData', "Data"), BYTES(8), NULL)

so we only need to use the ID macro when we need to keep the old camelName.

I'm not married to the word ID as the macro name, CAMEL would do as well.

We should also add a `.git/hooks/pre-commit` or `.git/hooks/pre-push` hook that verifies that no ids have been changed since the last version tag.

analyzer/analyzer.c
```diff
@@ -1241,7 +1243,16 @@ extern bool printFields(const Pgn *pgn, const uint8_t *data, int length, bool sh
       break;
     }
 
-    strcpy(fieldName, field->camelName ? field->camelName : field->name);
+    if ( showCamel )
```
meaning-matters, line 1246: `if (showCamel)` is consistent with existing code.

analyzer/analyzer.c
```diff
@@ -181,6 +182,7 @@ int main(int argc, char **argv)
     }
     else if (strcasecmp(av[1], "-camel") == 0)
     {
+      showCamel = true;
       camelCase(false);
```
meaning-matters, line 186: Since the term 'camel case' is already used (and it being the proper term), it should be used everywhere. Introducing a second term 'camel' for the same aspect/thing/... is bad. Also 'show camel' sounds weird.

sbender9: This is actually derived from the argument name, which is `-camel`

sbender9:
> How about using a nested macro call:
> 
> ```
>  #define ID(camel, nam) nam, .camelName = camel
>  ...
>  BINARY_FIELD(ID("someData', "Data"), BYTES(8), NULL)
> ```
> 
> so we only need to use the ID macro when we need to keep the old camelName.
> 
> I'm not married to the word ID as the macro name, CAMEL would do as well.
> 
> We should also add a `.git/hooks/pre-commit` or `.git/hooks/pre-push` hook that verifies that no ids have been changed since the last version tag.

Works great for BINARY_FIELD, but not for UINT8_FIELD:

```{UINT8_FIELD(CAMEL("someSid", "SID")), DISTANCE_FIX16_CM_FIELD("Heave", NULL), RESERVED_FIELD(BYTES(5)), END_OF_FIELDS},```

```
In file included from ./analyzer.h:66:
./pgn.h:2984:7: error: too many arguments provided to function-like macro invocation
 2984 |      {UINT8_FIELD(CAMEL("someSid", "SID")), DISTANCE_FIX16_CM_FIELD("Heave", NULL), RESERVED_FIELD(BYTES(5)), END_OF_FIELDS},
      |       ^
./pgn.h:519:48: note: expanded from macro 'UINT8_FIELD'
  519 | #define UINT8_FIELD(nam) UINT8_DESC_FIELD(nam, NULL)
      |                                                ^
./pgn.h:512:9: note: macro 'UINT8_DESC_FIELD' defined here
  512 | #define UINT8_DESC_FIELD(nam, desc) {.name = nam, .size = BYTES(1), .resolution = 1, .fieldType = "UINT8", .description = desc}
      |         ^
./pgn.h:2984:7: error: use of undeclared identifier 'UINT8_DESC_FIELD'
 2984 |      {UINT8_FIELD(CAMEL("someSid", "SID")), DISTANCE_FIX16_CM_FIELD("Heave", NULL), RESERVED_FIELD(BYTES(5)), END_OF_FIELDS},
      |       ^
./pgn.h:519:26: note: expanded from macro 'UINT8_FIELD'
  519 | #define UINT8_FIELD(nam) UINT8_DESC_FIELD(nam, NULL)
      |                          ^
./pgn.h:7535:29: error: invalid application of 'sizeof' to an incomplete type 'Pgn[]' (aka 'struct Pgn[]')
 7535 | const size_t pgnListSize  = ARRAY_SIZE(pgnList);
      |                             ^~~~~~~~~~~~~~~~~~~
../common/common.h:188:30: note: expanded from macro 'ARRAY_SIZE'
  188 | #define ARRAY_SIZE(x) (sizeof(x) / sizeof(x[0]))
```

analyzer/analyzer.c
keesverruijt: there is a .clang-format file in the repo, just run `make format` with clang-format installed to fix whitespace and other formatting.

keesverruijt:
Solution found, the problem with 

     UINT8_FIELD(CAMEL("camelCase", "name))
     
would be that CAMEL would be expanded first, so you get

    UINT8_FIELD("name", .camelName = "camelCase")

which obviously won't work. There is a C preprocessor trick to defer this one stage:

    #define EMPTY()
    #define DEFER(x) x EMPTY()

    #define CAMEL_NAME(camel, nam) nam, .camelName = camel
    #define CAMEL(camel, nam) DEFER(CAMEL_NAME)(camel, nam)

In the last push I've used a similar macro to define primary keys more naturally.

Unfortunately this only works for one level, so we will have to also define a `PK_CAMEL` as `PK(CAMEL(...))` or `CAMEL(PK(...))` won't work.

keesverruijt:
No, we don't need this for 6.0 but we do need it for the next one as I am now stating in the changelog that from now on no changes to Id fields will be made.

keesverruijt:
This creates macros PK() and CAMEL() that can be used simultaneously, if necessary.

PR #537, opened 2025-05-23
PGN 127506: Add primary key to "DC Type" field

tomer-w:
PGN 127506: Add primary key to "DC Type" field. I verified this field is rotating between various sources and provides values for each of them.


keesverruijt:
fixed in #541. cannot merge this PR as PK is now a macro, not a different field name.

PR #538, opened 2025-05-23
PGN 60928 - ISO Address Claim: add primary key to ISO Identity Number

tomer-w:
The ISO Identity Number field is constant per device and therefore should be marked as primary key.


keesverruijt:
Actually, it's the other way around: the other NAME fields should lose the primary key attribute. 

There is an _implicit_ primary key field which is the source address. We only need to add primary key attributes if the same PGN is sent by the same device (source address) with different content that is logically different from other content.

PR #546, opened 2025-06-17
SeaStar Helm PGNs

TamaGO89:
**WARNING** : I updated package.json to create a forked version of the package at @tamago89/pgns. I don't know how to do a pull-request for this without my last commit! please, be aware of it!!

I added support for the SeaStar rudder wheel.

- **65360 SeaStar: helm command** (replaced **Seatalk: Pilot Locked Heading** ): allows to dynamically change the resistance of the wheel and the port and starboard side limits.
- **65365 SeaStar: Helm Feedback** : the feedback of the wheel, with some warning flags, the ID of the wheel (up to 3 wheels are supported), encoder absolute position
- **65366 SeaStar: Helm Feedback** : published by the main wheel, contains detailed description of the wheels.

These messages are meant for a private N2K network for a Seastar compatible steering system.
I also updated the lookup table ( **SEASTAR_HELM_COMMAND** ) and an additional fieldtype ( **ANGLE_FIX16_CDEG** ).

I would like to also know a thing:
i wanted to publish a fork of the @canboat/pgns package but i don't know how to do it:

- if I simply publish the package i only get a package.json inside.
- if i run make generated on /analyzer i get an updated canboat.json inside /docs but i don't know how to link it to my node.js project.

Could you help me out understanding how to do this?
I've no experience with npm and very little with github, i did my research but i couldn't find anything useful.
Thank you!

### DIFF
--- a/analyzer/fieldtype.h
+++ b/analyzer/fieldtype.h
@@ -687,6 +687,13 @@ FieldType fieldTypeList[] = {
      .physical      = &ANGLE,
      .baseFieldType = "FIX16"},
 
+    {.name          = "ANGLE_FIX16_CDEG",
+     .description   = "Angle",
+     .resolution    = 0.01,
+     .unit          = "deg",
+     .physical      = &ANGLE,
+     .baseFieldType = "FIX16"},
+
     {.name = "ANGLE_UFIX16", .description = "Angle", .resolution = 0.0001, .physical = &ANGLE, .baseFieldType = "UFIX16"},
 
     {.name        = "GEO_DELTA_FIX24",
--- a/analyzer/lookup.h
+++ b/analyzer/lookup.h
@@ -2871,6 +2871,12 @@ LOOKUP(ZONE_SIZE, 4, "5 nm")
 LOOKUP(ZONE_SIZE, 5, "6 nm")
 LOOKUP_END
 
+LOOKUP_TYPE(SEASTAR_HELM_COMMAND, 4)
+LOOKUP(SEASTAR_HELM_COMMAND, 0, "No command")
+LOOKUP(SEASTAR_HELM_COMMAND, 1, "Power down")
+LOOKUP(SEASTAR_HELM_COMMAND, 2, "Reset")
+LOOKUP_END
+
 // Keep this at the end, so a next include does not need to do this
 #undef LOOKUP_TYPE
 #undef LOOKUP_TYPE_TRIPLET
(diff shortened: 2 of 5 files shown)


keesverruijt:
Hi Andrea,

I suggest you just remove `analyse/package.json` from your pull request.

Also, you need to fix some other stuff:

- instead of `MANUFACTURER_FIELDS` you should use `COMPANY(xxxx)` where xxxx is the numerical code for SeaStar (I think probably 1850?) this way the SeaStar PGNs are unique to SeaStar and they do not interfere with other brands.
- range -3060 to +3060 should probably be -3600 to +3600. But there is no need to show this in the description you can encode rangeMin and rangeMax in the field type definition, just add a few more of those.

Also would be good if you have an extraction of a logfile that shows these PGNs in action and add this to `samples/`.

keesverruijt:
Maybe somebody else can chime in on NPM, I have zero knowledge of it.

sbender9:
Please don't merge this @keesverruijt !

keesverruijt:
I have no intention to merge this, it is changing things that shouldn’t be changed and my original concerns have not been addressed. 

Maybe the author should join the Signal K discord and explain what he is trying to do there, then we can help him more interactively.

TamaGO89:
I'm sorry, i updated my fork without thinking that my commit would also go into this pull request.
My last commit is unrelated to this request.
I'll close this since the custom PGN that I need conflicts with the custom PGN that you need. So thanks for the explaination, since it's clear now that my branch would conflict with your needs.

Anyway (and you can skip this part, but i think it's due to explain what i did):
I work with canboatjs, there the PGNs are parsed using the "name" attribute as dictionary keys, which usually have special characters, white spaces, ... (since they match the field names inside the official documentation).
To parse data between PGN and Protobuffer, i need attributes to have alphanumerical names without whitespaces, so I opted for the "id" field, which is automatically generated from the "name" attribute.
This means that "PGN" becomes "pgn", which conflicts with the "pgn" field set by canboatjs.
I opted for the less intrusive solution (for my scenario), by renaming "PGN" into "PGN Data" (which "id" is "pgnData").

Again, i'm sorry, i wasn't thinking that this would automatically go into my pull request.

sbender9:
@TamaGO89 canboatjs 3.0 has support for using the Id field instead of Name. (camel case)

sbender9:
Let's close this until @TamaGO89 is ready?

PR #556, opened 2025-07-11
Fusion updates

sbender9:


sbender9:
@tomer-w can you review this? Not sure why there was a Power FUSION_MESSAGE_ID=32, I have tested with 28 and it works.

Also you look to see what the message is to set the BlueTooth device?
I use this: 2025-06-20T19:33:12.240Z,7,126720,0,255,11,a3,99,09,00,0b,%**s**,00,00,00,02,02

tomer-w:
Hi @sbender9, long time...
I can guess (with high confidence) why it was 32. The people who originally added the Fusion data assumed that the values used to report changes and to set changes are the same. This is not the case. For that reason, I split up those 2 enums: FUSION_STATUS_MESSAGE_ID which is sued when fusion want to report status (and the power status is really in value 32) and the original FUSION_MESSAGE_ID which supposed to be used for asking Fusion to change its status. As in my work I focused more on the first one, I didn't touch the original enum and relevant messages. I am aware that there is lot of mistakes there as people kept changing the values as they confused the set and get operations. Now it should be better as I defined two different enums.
Now moving forward, your change is correct based on my understanding of their work. I actually have much more data so I can fix it completely and remove all the PACKET_INCOMPLETE parts. The only problem I didnt do it now, as I was focused on other project (Victron stuff) and did not have enough quality time in my boat to test it. You can submit what you already did, and I can fix it further if you are willing to do some testing to my work. It should not take more than couple of hours and I think we should do it for the sake of the community.
Let me know how you want to proceed.

sbender9:
Thanks. I can definitely review and test any work you do.

tomer-w (approved):

sbender9:
now I am seeing that a ton of messages got removed, don't know how I missed that...

"Fusion: AM/FM Station" and many others

sbender9:
Go ahead and do a PR with your changes. I don't think ours conflict. I can merge them and test here.

@keesverruijt you can merge this PR when you have time.

tomer-w:
Opened: https://github.com/canboat/canboat/pull/558

PR #586, opened 2025-08-28
add Seatalk i70s brightness and color message definitions

fschaper:
feat(n2k): add Raymarine i70s Display Brightness and Color PGN definitions

- Added proprietary PGN 61439 (0xEFFF) decoding for Raymarine i70s display brightness
  * exposes raw brightness byte (0x00 … 0x4D observed) - might actually be percentage based, need to verify
- Added proprietary PGN 126720 decoding for Raymarine i70s display color scheme
  * supports Tag 1, Tag 2, Inverse, Red/Black palettes (as already defined by SEATALK_DISPLAY_COLOR)
- Based on reverse engineered CAN traffic from i70s

Analyzer now reports human-readable brightness and color selections
instead of generic proprietary data frames.

### DIFF
--- a/analyzer/pgn.h
+++ b/analyzer/pgn.h
@@ -1236,6 +1236,39 @@ Pgn pgnList[] = {
      PACKET_SINGLE,
      {COMPANY(358), UINT16_FIELD("Register Id"), SIMPLE_FIELD("Payload", BYTES(4)), END_OF_FIELDS}}
 
+    ,
+    {"Seatalk: Display Brightness (i70s)",
+     61439,                         // 0xEFFF (addressed, fast-packet)
+     PACKET_INCOMPLETE,
+     PACKET_FAST,
+     {COMPANY(1851),               // Raymarine
+      MATCH_LOOKUP_FIELD(PK("Proprietary ID"), BYTES(1), 140, SEATALK_MESSAGE_ID),
+      MATCH_LOOKUP_FIELD(PK("command1"), BYTES(1), 12, SEATALK_COMMAND),
+      LOOKUP_FIELD("Group", BYTES(1), SEATALK_NETWORK_GROUP),
+      BINARY_FIELD("Unknown 1", BYTES(1), NULL),        // observed constant 0x04 in your logs
+      MATCH_FIELD("Command", BYTES(1), 0, "Brightness"),
+      // Second FP frame starts here:
+      UINT8_FIELD("Brightness (raw)"),                  // 0x00..~0x4D (device-defined scale) - check if can be replaced with PERCENTAGE_U8_FIELD
+      BINARY_FIELD("Unknown 2", BYTES(1), NULL),        // observed 0x00
+      BINARY_FIELD("Reserved", BYTES(4), NULL),         // observed 0xFF 0xFF 0xFF 0xFF
+      END_OF_FIELDS } }
+
+    ,
+    {"Seatalk: Display Color (i70s)",
+     61439,
+     PACKET_INCOMPLETE,
+     PACKET_FAST,
+     { COMPANY(1851),
+       MATCH_LOOKUP_FIELD(PK("Proprietary ID"), BYTES(1), 140, SEATALK_MESSAGE_ID),
+       MATCH_LOOKUP_FIELD(PK("command1"), BYTES(1), 12, SEATALK_COMMAND),
+       LOOKUP_FIELD("Group", BYTES(1), SEATALK_NETWORK_GROUP),
+       BINARY_FIELD("Unknown 1", BYTES(1), NULL),        // observed 0x04
+       MATCH_FIELD("Command", BYTES(1), 1, "Color"),
+       LOOKUP_FIELD("Color", BYTES(1), SEATALK_DISPLAY_COLOR), // 0x01=Day1, 0x02=Day2, 0x03=Invert, 0x04=Red/Black
+       BINARY_FIELD("Unknown 2", BYTES(1), NULL),        // observed 0x00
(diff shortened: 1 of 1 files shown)


fschaper:
since both brightness and color share the same PGN 61439, one could also merge them into a single definition with a LOOKUP_FIELD("Command", BYTES(1), SEATALK_COMMAND) and a SWITCH_FIELD(...). Let me know what you prefer.

fschaper:
I don’t see any 126720 messages from my setup - only 61439 (0xEFFF) fast-packet messages with Proprietary ID=140, command1=12. I don't have my full logs at the moment but bay be able to double check on the weekend. 

However: So while analyzer/pgn.h already has entries for Seatalk1: Display Brightness/Color (PGN 126720), what my i70s actually puts on the bus are addressed fast-packets with PGN 61439. For example, these two frames will switch the i70s on CAN ID 6 to the Red/Black scheme:

cansend can0 1DEFFF06#20093B9F8C0C0401
cansend can0 1DEFFF06#21010300FFFFFFFF

That’s why I added new definitions for 61439 - so analyzer shows those messages decoded instead of just “Manufacturer Proprietary.”

My original goal here is to sync the i70s palette with an Axiom Pro 2 S day/night palette change. So whenever the MFD toggles, I will replay the corresponding brightness/color command back onto the bus for the i70s's.

sbender9:
I have a Signal K plugin that's syncs displays of different vendors. 

I never found the Axiom to send out the day/night change on the network. Did you find something?

sbender9:
Do you have your i70s set the share brightness? If so, you she see the existing pgns being sent.

fschaper:
> And FYI, that plugin works with the i70s

Great to hear, and thanks again. I am already a happy user of some of your Signal K plugins. Thank you for your hard work!

So if the i70s responds fine, that suggests Raymarine have kept some backwards compatibility in place. The display still reacts to the older PGN definitions, but the devices themselves are now emitting the newer PGNs. That feels a bit odd because it would imply mixed formats on the bus and in theory that could break palette sync with older equipment unless they really do listen for both.

fschaper:
> Do you have your i70s set the share brightness? If so, you should see the existing pgns being sent.

I’ll try to verify again that there really are no PGN 126720 messages being sent on my system. So far I’ve only seen PGN 61439 fast-packet frames when changing brightness or color, never 126720. I'll happy to do/provide some logging.

For context, my setup looks like this:
- 4× i70s
- one i70s in its own group
- three i70s in a shared group together with 2× Axiom Pro 2 S and a p70s
- brightness sharing enabled on both groups

fschaper:
Great, sounds like we’ll both be on the boat this weekend then. As long as the family will bear with me I’ll double-check for 126720 on my setup and share whatever I find.

If time allows I’ll also try to trace the PGNs for switching the EV-2/EV-200/p70s autopilot between Auto and Standby - that’s still giving me issues with your SignalK autopilot plugin. Having raw captures for that would be useful too.

If there’s anything specific you’d like me to try or log while I’m out, just let me know. I’m happy to contribute additional captures or tests.

sbender9:
Have you checked out visual-analyzer? Big help looking at stuff like this. 

I'm really surprised those don't use the same pgns as the EV-1/ACU-100. 

Definitely take a look with visual-analyzer and get some logs of changing the AP mode and heading.

fschaper:
Heading works fine on my setup, it has only been stubborn with toggling between Auto and Standby. I will use visual-analyzer when I am back on the boat this weekend and capture logs of the mode changes so we can see exactly what is on the bus. Thanks for the tip.

sbender9:
I do not see 61439 from my i70s. I think I may be one software update behind...

sbender9:
Can you get me a log of the PGN 61439 data?

fschaper:
> Can you get me a log of the PGN 61439 data?

Sure, I’m on the boat now and will be able to provide a log later today.
Would you prefer that I apply a device filter (e.g. can device 6) to keep it focused, or would a more verbose capture be more helpful?

sbender9:
Verbose is fine. I can filter here

sbender9:
In your description, you say "Added proprietary PGN 126720", but there is no related change in the PR.

fschaper:
> In your description, you say "Added proprietary PGN 126720", but there is no related change in the PR.

You are right, there is no change to 126720 here. That reference was left over from when I was still reverse engineering the packets. This PR only adds decoding for 61439, which is what my i70s actually sends. Will follow up with the logs.

sbender9:
I updated the software on the i70s, now I am getting 61439

sbender9:
I don't need a log now

fschaper:
> I don't need a log now

Interesting. Did you check whether both PGNs are active in parallel or if the device has switched over completely?

sbender9:
And I spoke to soon, it is not sending 61439. I am getting 61184 and 130847, which are new.

fschaper:
[can0_log_i70s_invert_to_rb_to_invert.log.zip](https://github.com/user-attachments/files/22060071/can0_log_i70s_invert_to_rb_to_invert.log.zip)

and no, `candump -L can0,19EFFF00:1FFFFF00` gave me nothing

fschaper:
I'm at firmware version 3.24 for the i70s

sbender9:
Same here now.

keesverruijt:
Anyway, I will let you guys get to the bottom of this before I merge…

fschaper:
I think I found my mistake: I interpreted the CAN-ID as a broadcast (PDU2) message. The correct PGN is 126720 because it's a PDU1 (addressed) message, and the PDU Specific (PS) byte is ignored in the calculation.

fschaper:
Learned something valuable about the N2K standard in the process - sorry for taking your time!

Issue #309, opened 2022-10-05
Reverse engineering of Navico autopilot commissioning


keesverruijt:
(A lot of) work has already been done on this by @htool, see https://github.com/htool/RaymarineAPtoFakeNavicoAutoPilot 

Hans' focus was to be able to use a Raymarine AP with a Navico control head, for this to work it needs to know that the AP is fully commissioned. 

I intend to go the other way:

* Add PGNs for the most common reports like mode (Standby/Auto/Wind/FollowUp).
* Add PGNs for the most common commands like mode / -10 / -1 / +1 / +10.
* Add PGNs that make the control head believe the autopilot is commissioned (if this exists, I think this may not be possible for all control heads.)

TwoCanPlugIn:
Was following up a few issues I had raised on SignalK and saw a link to this. Canboat has been an inspiration for the OpenCPN TwoCan plugin, so happy to give something back.

I haven't been able to get the NAC-3 to go into Navigation or No-Drift Mode because I have yet to figure out what other PGN's are required. For example, in Navigation Mode, even though I generate and transmit valid PGN 129283 (XTE), PGN 129284 (Navigation) and PGN 129285 (Route/Waypoint List) messages, they are not sufficient to enable the NAC-3 to engage in Navigation Mode. There are a couple of PGN's sent by a B&G Zeus MFD that appear to be necessary to support Navigation Mode, but as yet, I have not been able to decipher them

Note the NAC-3 behaves as four separate NMEA 2000 devices, Autopilot, NMEA 183 Input, Rudder Feedback and Virtual Rudder Feedback. Similarly the Zeus-2 MFD behaves as several NMEA 2000 devices, MFD, Navigator, Autopilot Controller, GPS.

All of these are Manufacturer Proprietary PGN's, with the first two bytes encoded with the Manufacturer and Industry codes as per usual.

PGN 65305, 
Broadcast by NAC-3 Autopilot. 
Reports the Status and Mode of the Autopilot
These correspond to what is displayed on the Zeus-2 MFD

NAC-3 Autopilot Status 

Sample Raw Data
0x41 0x9f 0x00 0x02 0x10 0x00 0x00 0x00
0x41 0x9f 0x00 0x02 0x02 0x00 0x00 0x00
0x41 0x9F 0x64 0x02 0x10 0x00 0x00 0x00

Byte 2 indicates the Autopilot Model. 0x64 - NAC-3, 0x00 - AC-12
[...]

PR #593, opened 2025-09-30
Improve Simnet Autopilot Support

sbender9:
###  I don't love that are separate entries for different modes, but depending on the event, the rest of the bytes have different meanings.

Would be cool I guess if we could specify multiple values for a match? Any another ideas.

###  The Change Course one is a bit strange and have not really figured it out.

I see the following when do +/- buttons:

'+1': '03,AE,00',
'+10': '03,D1,06',
'-1': '02,AE,00',
'-10': '02,D1,06'

The first byte is direction, port or starboard. I can't make any sense out of the second two bytes. Anyone have any Ideas?

###  I have not figured out how to directly set the target heading or wind direction, or if it is even possible

#309 

BTW, The signal-autopilot plugin now supports controlling Simrad autopilots.I could use testers there, I have only testes with my NAC-3.


keesverruijt (changes requested):
Some things to improve I think.

keesverruijt:
yup. I will try tomorrow otherwise Friday.

PR #601, opened 2026-04-02
Adding python projects based on canboat to the Related Projects section

tomer-w:
Adding couple useful projects which are based on this amazing effort. It will be nice to include them in your list.


tomer-w:
Thanks!!

PR #624, opened 2026-05-01
Garmin SteadyCast PGNs

negrusti:
Garmin SteadyCast PGNs


negrusti:
@keesverruijt @sbender9 the CI/CD is failing with Segfault:
Replacing schema version in ../docs/canboat.xsd with 2.3.0
../rel/linux-x86_64/analyzer-explain -explain-xml -camel >../docs/canboat.xml
Segmentation fault (core dumped)

No idea how to fix this. Am I missing some fields?

sbender9:
I would start by adding just on PGN at a time to try to narrow it down.

Then go field by field.

negrusti:
> I would start by adding just on PGN at a time to try to narrow it down.
> 
> Then go field by field

All of them have the same structure. But I can't operate on Segmentation fault :)

PR #625, opened 2026-05-02
fix: update stale external documentation URLs

sentientapp:
## Summary

- Replace stale external documentation links with working current URLs or web archive snapshots
- Regenerate generated docs from the updated metadata

## Notes

- No useful snapshot was found for `http://askjackrabbit.typepad.com/ask_jack_rabbit/page/7/`, so the PGN 130306 reference now points to the archived Maretron WSO100 User Manual, which documents that PGN.
- `docs/canboat.xsl` now links to the newer NMEA 2000 PGN Field List, version 3.002 from 2024, instead of the older 1.301 list from 2010.
- A few source links remain broken because no valid replacement or useful web archive snapshot was found:
  - http://www.nmea.org/Assets/20190624%20NMEA%20Heartbeat%20Information%20Amendment%20AT%2020190623HB.pdf
  - https://www.nmea.org/Assets/20190613%20windlass%20amendment,%20128776,%20128777,%20128778.pdf


sentientapp:
One note on the ITU M.1371-5 references: I replaced the broken URL with an archived copy of the same recommendation page URL:

`https://web.archive.org/web/20251212005646/www.itu.int/rec/R-REC-M.1371-5-201402-I/en`

That archived page is a language and download format selection page rather than a direct PDF. I am not sure whether the original `/en` URL was intended to behave that way or to go directly to the English document.

If we prefer a direct document link, the archived English PDF is available here:

`https://web.archive.org/web/20251213141920/www.itu.int/dms_pubrec/itu-r/rec/m/R-REC-M.1371-5-201402-I!!PDF-E.pdf`

I left the recommendation-page URL in this PR as the most minimal replacement for the broken original link.

Issue #573, opened 2025-07-20
Trigger FAST format in analyzer when used with *-serial


keesverruijt:
When not reading from a logfile but a pipeline fed by `antisense-serial` or `ikonvert-serial` the format is FAST, e.g. all data for fast pgns on one line; the tools could write a header that explains the format which would be understood by `analyzer` so it doesn't mistakenly see the first few short fast PGNs as unreadable.

See #569 for such an occasion where a user didn't realise this is what was going on.

PR #631, opened 2026-05-17
feat: add maretron-ipg, a TCP client for the Maretron IPG100 gateway

sentientapp:
## Summary

  Add `maretron-ipg`, a new bidirectional gateway tool for the Maretron IPG100
  NMEA 2000-to-Ethernet bridge. Slots in alongside `actisense-serial` and
  `ikonvert-serial` as another canboat tool that adapts one wire format to
  canboat plain-CSV.

 ## Details

  - This **does not consume** one of the IPG's licensed-client slots. Operators can 
    run multiple `maretron-ipg` connections alongside paid clients (N2KView).
  - The IPG100 caps total simultaneous client connections at 20 regardless of
    type, so `maretron-ipg` sessions still count toward that limit.
  - Binary mode is required: the IPG's default ASCII output is decoded using
    its internal PGN dictionary, which is limited.
  - The IPG100 does **no device-side PGN filtering** in either direction. Every
    frame on the bus reaches every connected client, and every frame any client
    writes is forwarded both to the bus and to all *other* connected clients.
  - The IPG handles fast-packet reassembly itself, so each frame carries a full
    logical N2K payload.
  - On TX, source address on the wire is always `0xFF` — the IPG substitutes
    its own claimed SA.

  ## Test plan

  - Smoke-tested both RX and TX against a Maretron IPG100. Receive
    decodes through `analyzer -json -nv` to the expected JSON for live bus
    traffic; transmit verified end-to-end by a separate N2K logging device
    on the same bus capturing every frame the IPG forwards.


sbender9:
Hats off to your AI. TLDR. Will canboatjs need changes for this?

sentientapp:
I slimmed down the description.  This works fine with canboatjs if the `-q` flag is used to suppress INFO logs and the `toChildProcess: "nmea2000out"` option is used to enable outbound writes.

I do get an `Error: Parser not found for input. - # format=FAST` thrown on startup from `parseN2kString` due to #573.  Same issue is currently affecting `actisense-serial` and `ikonvert-serial`.  Submitted canboat/canboatjs#429 to fix it.

I have a TypeScript version of `maretron-ipg` ready to PR later today.

keesverruijt:
Sorry, the mail hit my junk folder and I am crossing the pacific right now!

It's amazing what the AIs can do nowadays. Happy to merge.

sentientapp:
Thanks for the merge, Kees!  Safe travels.  Hope the weather's cooperating.

PR #637, opened 2026-06-07
fix(130820): BEP CZone alarm string response (builds on #636)

dirkwa:
## What

Defines **PGN 130820** for BEP Marine / CZone (MFG 295): renames `BEP Marine: CZone Enumeration Reply` → **`BEP Marine: CZone Alarm String Response`** and gives it `Device ID` / `Channel` / `String` fields.

## Relationship to #636

This builds on **#636** (which defines PGN 65299 as `BEP Marine: CZone Alarm String Request`). **#636 should merge first.** PGN 130820 is the *response* to the 65299 request: the device sends 65299 asking for an alarm's text, and the addressed module replies with 130820 carrying the matching string.

The field names here intentionally mirror #636's 65299: the `Device ID` and `Channel` in 130820 are echoed back from the triggering 65299 request, so a consumer can correlate the response to its request.

This PR no longer touches 65299 — that PGN is defined by #636. The two PRs are otherwise on disjoint PGNs and don't conflict.

## Fields (130820)

| field | wire offset | type | notes |
|---|---|---|---|
| Device ID | byte 2 | uint8 | echoed from the 65299 request |
| Channel | bytes 3-4 | uint16 LE | channel / alarm id, echoed from the request |
| String | bytes 5.. | NUL-terminated string | the alarm string text (ASCII/UTF-8 not yet confirmed) |

Links to the CZone protocol notes at https://github.com/dirkwa/czone-spec for the request/response details.

## Notes

- 130820 is `PACKET_INCOMPLETE` (fast packet); the string is variable-length and NUL-terminated.
- Generated `docs/canboat.{json,xml,html}` regenerated via `make generated`; the diff is scoped to the 130820 entry only.

## Tested

- `make` builds clean.
- `make generated` runs clean; the regenerated `docs/` diff contains only the 130820 entry.


dirkwa:
I've rescoped this to PGN 130820 only, building on top of #636. Since #636 defines 65299 as the CZone alarm string *request*, this PR now covers 130820 as the *response* — echoing the same Device ID / Channel back, followed by the alarm string. I dropped my earlier 65299 change so the two don't overlap; #636 should merge first. The field names here mirror #636's so the request/response pair reads consistently.

PR #641, opened 2026-06-08
PGN 65284, 65301 - BEP Marine: CZone Circuit State. Evinrude, Mercury, Yanmar PGN stubs

negrusti:


analyzer/lookup.h
```diff
@@ -404,6 +404,61 @@ LOOKUP(CZONE_ALARM_TYPE, 63, "Battery Temperature High")
...
+LOOKUP(CZONE_MODULE_TYPE, 16, "Display Interface (DI)")
+LOOKUP(CZONE_MODULE_TYPE, 17, "Wireless Interface (WI)")
+LOOKUP(CZONE_MODULE_TYPE, 28, "Control 1")
+LOOKUP(CZONE_MODULE_TYPE, 29, "Keypad")
+LOOKUP(CZONE_MODULE_TYPE, 30, "Contact 6 P")
+LOOKUP(CZONE_MODULE_TYPE, 31, "Combination Output Interface (COI)")
+LOOKUP(CZONE_MODULE_TYPE, 54, "CXP")
+LOOKUP_END
+
+LOOKUP_TYPE(CZONE_EXTENDED_STATE_FLAG_LAYOUT, 5)
```
keesverruijt, line 427: `CZONE_EXTENDED_STATE_FLAG_LAYOUT` is defined here (32 entries) but never referenced anywhere in `pgn.h` — it's dead code. Its content is instead duplicated as prose in the 65301 `Circuit Flags` description. Either wire it into a `LOOKUP_FIELD` or drop the lookup and keep just the prose.

keesverruijt (commented):
Thanks for this. CI is green and it rebases cleanly on top of the 6.2.1 BEP/CZone work. A few things to resolve before this can go in.

**1. The CZone 65284/65301 reinterpretation overlaps with what just shipped in 6.2.1.** Those PGNs were defined in #637/#638 (dirkwa) with deliberately neutral, "purpose unverified" field names and `.url` links to the czone-spec. This PR overwrites that with confident semantic names (`Module Type` lookup, `Page`, `Circuit Flags`), flips both to `PACKET_COMPLETE`, and drops the `.url` references. Before we replace the just-released interpretation, can you share where the `CZONE_MODULE_TYPE` ids and the channel/page-layout mapping come from (a capture, a spec, a device)? If it's solid, great — but please keep the czone-spec `.url` links.

**2. The 14 Evinrude/Mercury/Yanmar additions are unsourced stubs.** Each is a bare `BINARY_FIELD("Data", ...)` placeholder with a speculative name and no `.explanation` or `.url`. canboat generally wants either a sample capture or a sourced reference before naming a proprietary PGN. Could you split these into their own PR and add provenance for the names? It also keeps the CZone semantic work reviewable on its own.

**3. One concrete bug:** the new `CZONE_EXTENDED_STATE_FLAG_LAYOUT` lookup is defined but never referenced (dead code) — see inline comment.

Inline notes below.

analyzer/pgn.h
```diff
@@ -2119,17 +2163,25 @@ Pgn pgnList[] = {
                     "The Message Type field selects the variant."}
 
     ,
-    {"BEP Marine: CZone 65301",
+    {"BEP Marine: CZone Circuit Extended State",
```
keesverruijt, line 2166: This replaces the neutral `BEP Marine: CZone 65301` definition just released in 6.2.1 (Field 1/2/3 + Status Bitmap, "purpose unverified", with a czone-spec `.url`). Flipping to `PACKET_COMPLETE` with semantic names asserts the layout is verified — what's the source for the `Module Type` ids and the 32/16/8-channel page mapping? Also please don't drop the `.url` link.

analyzer/pgn.h
```diff
@@ -1779,22 +1793,24 @@ Pgn pgnList[] = {
      {COMPANY(175), BINARY_FIELD("Data", BYTES(6), ""), END_OF_FIELDS}}
 
     ,
-    {"BEP Marine: CZone Circuit Status",
+    {"BEP Marine: CZone Circuit State Bitmap",
```
keesverruijt, line 1796: Same concern as 65301: this rewrites the 6.2.1 `CZone Circuit Status` definition and flips it to `PACKET_COMPLETE`. Keep the czone-spec `.url` and please source the `Module Type` interpretation.

analyzer/pgn.h
```diff
@@ -1730,6 +1730,13 @@ Pgn pgnList[] = {
                     "classifier; its per-source semantics are not yet observed in real captures (likely a sub-type / "
                     "channel index, an alarm-state polarity flag, or a frame counter)."}
 
+    ,
+    {"Evinrude: Joystick Control Primary",
```
keesverruijt, line 1734: These Evinrude/Mercury/Yanmar entries (14 in total across the PR) are bare `Data` stubs with no `.explanation` or `.url`. Where do the names come from? Better as a separate PR with a sample or source for each name.

keesverruijt:
These can no longer be merged, as the .h files are no longer the source of truth.

Also we no longer accept stubs that do not provide any actual fields etc.

Issue #591, opened 2025-09-03
Why reserve 3 bits for values with bitlengths >= 8?


inquestruss:
Most of the comments and documentation say that fields with bitlengths >=4 reserve two values: maximum positive value for missing, maximum positive value - 1 for invalid.

This code (fieldtype.c, getMaxRange):
  uint64_t specialvalues = (size >= 8) ? 3 : (size >= 4) ? 2 : (size >= 2) ? 1 : 0;

reserves a third value.

keesverruijt:
This is a good point, the exact history of this evades me. I think it was an old (< 2010) document by the NMEA when they still leaked stuff that indicated this.

Let me investigate this, but TBH this has low priority (as evidenced by my slow response...)

keesverruijt:
Well on reading the spec it turns out that three special values is correct. Now if somebody can enlighten us on what the third value means would be even better. See #671

keesverruijt:
## Resolved: the 3rd reserved value is spec-accurate — it means "Reserved"

I went to the open sources to settle what that 3rd special value (MAX−2) actually is. The **primary source** is Frank Cassidy's *"NMEA 2000 Explained — The Latest Word"* — and Cassidy is the same author who wrote the Appendix B PGN-table PDF, so this is as authoritative as public sources get. It states verbatim:

> uint16: 16-bit unsigned integer, range 0 to 65,532. **65,533 = Reserved. 65,534 = Out of range. 65,535 = Data not available.**

So for 8-bit-and-larger integers the top **three** values are reserved, by spec:

| value (uint8 / uint16) | NMEA term | our concept |
|---|---|---|
| MAX `0xFF` / `0xFFFF` | **Data not available** | Unknown |
| MAX−1 `0xFE` / `0xFFFE` | **Out of range** | Error |
| MAX−2 `0xFD` / `0xFFFD` | **Reserved** | *(not modeled)* |

Corroboration:
- The NMEA 2000 v1.300 Appendix B datatype dictionary lists `uint8 = "0 to 252"`, `uint16 = "0 to 65,532"`, `uint32 = "0 to 4,294,967,292"` — i.e. MAX−3 is the last valid value (3 reserved), consistent across **every** integer datatype.
[...]

PR #672, opened 2026-06-25
feat(540): expose Unknown/OutOfRange/Reserved special values per field

### LINKED ISSUE #540: Are we sure that 3 bits number has "unknown" value?
In my system I get PGN 60928 - ISO Address Claim which looks like this:
```2025-04-10T12:10:16.614Z,6,60928,5,255,8,f5,01,c0,2c,ef,aa,46,c0```
The output of this is:
```
2025-04-10T12:10:16.614Z 6   5 255  60928 ISO Address Claim:  Unique Number = 501; Manufacturer Code = Victron Energy; Device Instance Lower = Unknown; Device Instance Upper = 29; Device Function = Battery; Device Class = Electrical Generation; System Instance = 0; Industry Group = Marine; Arbitrary address capable = Yes
```
It all makes sense except the ```Device Instance Lower = Unknown``` part. Are we sure they meant it to be unknow? it makes more sense to be 7. I think that at least some 3 bits number do use all the bits.

What do the NMEA 2000 experts think about this example?

### PR DESCRIPTION
**Pull request by keesverruijt**
Closes #540.

Makes NMEA 2000's top-of-range non-data sentinels **explicit** in the XML/JSON for integer number fields, instead of leaving every consumer to recompute them from the bit width.

```xml
<RangeMax>252</RangeMax>            <!-- last valid value -->
<UnknownValue>255</UnknownValue>       <!-- data not available (most positive) -->
<OutOfRangeValue>254</OutOfRangeValue> <!-- out of range (max-1, >=4 bit)        -->
<ReservedValue>253</ReservedValue>     <!-- reserved (max-2, >=8 bit)           -->
```

Per **Cassidy, "NMEA 2000 Explained"** (the authoritative public source — Cassidy also wrote the Appendix B PGN-table PDF), 8-bit and larger integers reserve three values; 4-7 bit reserve two; 2-3 bit reserve one. See #591 for the supporting research and the durable archive link.

## What changed

- **`fieldtype.c`** — one helper (`reservedCountForSize`) now drives **both** the `RangeMax` and the sentinels, so they can't drift. A field reserves **none** when its range spans the full unsigned width (the "all values valid" idiom, e.g. ISO device instance — #540's original `Device Instance Lower = 7 = Unknown` complaint). The full-range check compares against the **unsigned** raw max so it's stable between the explain (SI/radians) and runtime (user/degrees) builds. New **`SPECIAL_VALUES(n, "Name")`** macro sets the count explicitly.
- **`print.c`** — decode now strips exactly `reservedCount` sentinels. Previously a separate heuristic stripped 2 and disagreed with the published `RangeMax` for ≥8-bit and signed fields (a signed `0x7FFF` could decode as a value exceeding the field's own `RangeMax`). Sentinel output renamed `ERROR` → **Out Of Range**, `RESERVED1` → **Reserved** (and the `DATAFIELD_*` macros).
- **`analyzer-explain.c` + `canboat.xsd`** — emit the three optional elements for plain numeric fields (<64 bit; not lookups/strings/reserved/float/ISO_NAME/match). `canboat2json.xslt` needs no change (generic numeric passthrough). 64-bit is excluded since `UINT64_MAX` exceeds JSON's safe integer range.
- **`SCHEMA_VERSION`** 2.3.0 → 2.4.0 (additive: new optional elements/keys).
- `AGENTS.md` documents the convention and `SPECIAL_VALUES`.

## Behaviour change (intended)

Decode now strips the spec's sentinels consistently: an 8-bit `253` prints `Reserved` (was the number), and signed sentinels like `0x7FFF` now decode to Unknown/null in **both** the plain and JSON paths (they previously disagreed). This fixes decoding "data not available" correctly.

## Verification

- `make generated` green: full golden suite (analyzer + n2kd) passes, schema + JSON-range validate, and the NMEA cross-check passes.
- **Generated docs diff is purely additive** — only the three new elements + the version bump, **zero `RangeMax` churn**.
- J1939 build compiles (shares the changed code).


keesverruijt:
@aisle-security check please

Issue #685, opened 2026-06-27
Decode 65280 proprietary engine PGNs (Mercury / Honda / Yanmar)


keesverruijt:
PGN **65280** (`0xFF00`) is a multiplexed single-frame proprietary PGN. There is evidence of dedicated per-manufacturer **engine** data variants:

- **Mercury** (mfr 144) engine data
- **Honda** engine data
- **Yanmar** (mfr 172) engine data

(The PowerPole anchor also uses this PGN, and **BEP** (mfr 295) circuit control is already covered as *BEP Marine: CZone Circuit Control*.)

These engine variants appear to be real, but their field layouts are not yet established well enough to define. **Real-world 65280 captures from each engine brand — with an address claim (60928) and product info (126996) in the same log — would be the most valuable input** for confirming the manufacturer keys and completing the layouts.

Split out from #679. Note: the **Navico** (mfr 275) 65280 is a display broadcast, not engine-related, and was dropped from #679 until more data is available.

keesverruijt:
### Field hint from the canboat-py extended decoder (#676)

The `canboat-py-gh` extended decoder (`nmea2000/semantics.py`) annotates the
per-source **"Run State"** byte of the engine 65280 as:

> Per-source run-state byte (enum, values **0 / 4 / 5 / 6**): a dynamic
> operational state that cycles through key-on / key-off transitions — **not** a
> static config or steady telemetry value.

So whichever brand's 65280 we decode here, expect one byte to be a small
run-state enum (0/4/5/6) tracking key-on/key-off, best modelled as a
`LOOKUP_FIELD` rather than a numeric telemetry field. This is consistent with
65280 being engine status rather than configuration, and corroborates dropping
the Navico (mfr 275) Triton *display* 65280 from #679 as non-engine.

Source: [`HyperActiveJ/canboat-py-gh`](https://github.com/HyperActiveJ/canboat-py-gh)
`boat-decodes`, `nmea2000/semantics.py` `ANNOTATIONS[(65280, "Run State")]`.

keesverruijt:
## Investigation: can we define definitive fields for the 65280 engine variants?

Looked into whether clear, definitive field layouts can be established for the manufacturer-proprietary 65280 engine variants. Summary of findings:

**Mercury (144) & Honda**
[...]

Issue #679, opened 2026-06-26
Decode additional Navico proprietary PGNs from #676 fork (65280, 65313, 130822, 130825)


keesverruijt:
From discussion #676 (fork [HyperActiveJ/canboat@6089f21](https://github.com/HyperActiveJ/canboat/commit/6089f210167e1fa70e8975948849d582ae09f636)).

Navico (mfg **275**):

- **65280 "Engine Identity/Config"** (single): Field A, Run State, Config Code.
- **65313 "Depth Quality"** (single) — replaces the current "Navico: Proprietary" stub: Instance, Depth Quality (FIX8, 0.01 signed, "0=no bottom lock .. ~1.0 good").
- **130825 "Alert"** (fast) — replaces "Navico: Unknown 2": Instance, Record ID, Alert Type, Alert ID, Alert State (ALERT_STATE), Action Flag, Alert Severity (ALERT_TYPE), Value. Pairs with Mercury 130822 sub-type-2 text records by Alert ID.
- **130822** — fork models sub-type 1 as a "Keyed Parameter Table" and sub-type 2 as an "Alert" numeric record.

⚠️ **Overlap / conflict:** the fork's 130822 interpretation conflicts with the source-selection decode on our local `navico_specific_pgns` branch (cmd-dispatch container + tDataType id space). These two readings of 130822 must be **reconciled** before either is merged. Tracking that here.

Source: unmerged fork; port + reconcile, build, verify.

keesverruijt:
Resolved by #695 (and #686). Navico proprietary PGNs from the #676 fork are now decoded: 65313 (#686), 130825 (#686/#695), and 130822 reframed as the UDB key/value database transport (#695). The remaining 65280 engine PGNs are tracked separately in #685.

PR #686, opened 2026-06-27
feat(navico): decode 65313 Depth Quality and 130825 Alert

keesverruijt:
From discussion #676 (HyperActiveJ fork). Adds Navico (mfr 275):

- **65313 "Depth Quality"** — replaces the unidentified "Navico: Proprietary" stub: `Instance` + a signed depth-quality ratio (0 = no bottom lock … ~1.0 = good).
- **130825 "Alert"** — replaces "Navico: Unknown 2": Alert Type/ID, Alert State lifecycle (`ALERT_STATE`) and Alert Severity (`ALERT_TYPE`).

Includes `samples/navico-proprietary.raw` (real frames from @HyperActiveJ; both PGNs decode).

### 65280 dropped
The fork also proposed a Navico (mfr 275) **65280 "Engine Identity/Config"**. It has been **dropped** — **we find no evidence the Navico-275 variant is engine-related**, and the fact that a **Triton (a display/instrument) generates it** makes an engine meaning extremely unlikely. Our reverse-engineering research backs this up: the *engine* 65280 decoders belong to **Mercury / Honda / Yanmar** (each under its own manufacturer code), while the only Navico-275 reference is a Triton *encoder* with no decoder or field semantics anywhere. The engine variants are tracked separately in #685. 65280 stays out until there's real data.

### 130822 deferred
The fork's 130822 "Keyed Parameter Table"/"Alert" decode is **not** included: it reinterprets the same discriminator byte that local source-selection work reads as a command; the two must be reconciled first.

Verified: `make` / `make generated` / `make tests` green.

Refs #679


pete-jorgensen:
Unclear where triton comes I to play?? in my case I have a mercury engine, which it appears claude found supporting device for??

keesverruijt:
I checked my RE and my Tritons generate this, so ...

keesverruijt:
Your Mercury engines generate it, but with a Mercury company ID, not Navico/Simrad/B&G I hope/think.

pete-jorgensen:
I'll cross check, mercury uses rebranded navico devices for some things ex the vessel view displays, so it gets muddled sometimes if it's mercury specific of navico generic

pete-jorgensen:
My AI says this:

**65280 "Engine Identity/Config" (Navico 275): drop it — our entry was the weak one.**
Its fields are `fieldA` / `runState` / `configCode`: placeholder names, no units, no
evidence. The canonical engine 65280s (Mercury 144, Honda 175, Yanmar 172) are conservative
raw `data` passthroughs, so ours actually over-claims by asserting engine semantics under a
manufacturer code that otherwise only appears as a Navico display. We won't repush without a
capture proving those bytes track engine identity.

**130822: one question settles the deferral.** Our three 275 variants dispatch on a single
discriminator — `subType` at **BitOffset 24 (data byte 3, right after the 2-byte
manufacturer/industry header)**:

| subType | variant | extra match |
|--------:|---------|-------------|
| 1 | Keyed Parameter Table | — |
| 2 | Alert | `recordKey` == 32792 |
| other | catch-all binary blob | — |

**Which byte offset does your source-selection/command reading consume?** If it's also
offset 24, your command-id and our `subType` are the same field — reconcile to one decode.
If it's a different offset, the two don't collide and the deferral can lift. I can share the
`navico-proprietary.raw` frames annotated with byte-3 values to confirm against real data.

keesverruijt:
I think those Mercury displays are OEM versions of the Simrad IS-42 and B&G Triton 2, so I'm not surprised actually. Still before we added it should be substantial. Any time we get the Id for a PGN wrong, we get bitten. And since the Id was (stupidly, stupidly) derived from the user viewable name, I tend to hold off until there is real proof.

What you don't know is that I am doing more RE work on 130822. I _think_ we made a real breakthrough. Will know in a day or so.

pete-jorgensen:
Understood.

pete-jorgensen:
Here's what I know about 130822 from my logs and what I'm planning to capture to close gaps:

# PGN 130822 — what we know

The key thing up front: **there are two completely distinct PGN 130822 messages**, and
both have their structure solved.

## Two variants

| Variant | Mfg | Header | On our bus? | Decode state |
|---|---|---|---|---|
| **Navico** | 275 | `0x1399` | **Yes** — in every engine log, emitted by the VVL gateway (src 2/8) | solved-core, conf **94** |
| **Mercury Marine** | 144 | mfg-144 trio | **No** — known structurally, never seen in the corpus | solved-core |

> Same PGN number, different manufacturers → entirely different layouts. canboat dispatches
> them by the manufacturer/industry Match fields. Only the **Navico** one matters for the
> boat's actual traffic.

## Navico 130822 (the one on the bus) — `analysis/pgn_overlay/130822_navico.json`

It's a **variable-format fast packet** from the VVL gateway. The structure is **fully
decoded and wire-confirmed against the corpus** (`scripts/decode_130822_subtypes.py`):

- **7-byte common head** (`13 99`, id@2, sub-type@3, u16 record-key@4:6, byte@6), then the
  body layout is selected by the **sub-type byte**. Every observed sub-type is decoded:
  - **Sub-type 1** — 15-byte keyed table; one parameter record per frame, key cycles over
    13 record IDs.
  - **Sub-type 2** — alert records (`0x8008` text / `0x8018` numeric) plus numeric variants
    (`0x8010` 28-byte, `0x8004` 37-byte) with trailing IEEE-754 LE floats.
  - **Sub-type 6** — `0x0900`, 49-byte composite: u16 value + u16 blob-length + signature +
    a variable nested TLV blob of `{tag, float}` groups.

**The sub-type-1 keys are reclassified into three tiers** (cross-session scan of all 20 logs):

- **2 fast live counters** `0x0900`/`0x8010` — cumulative **engine-life counters**
  (r≈0.998 with engine-hours, identical on src2/src8 → gateway-aggregated, flat within a
  session through RPM sweeps, monotone with engine-hours not calendar).
- **8 slow keys** (`0x0100`…`0x0800`, `0x8000`, `0x8004`) — additional slow engine-life
  parameters drifting in lock-step over 16 months.
- **3 truly-constant capability keys** `0x0600`/`0x8011`/`0x8014` — the genuine fixed
  capability/limit constants.

So **11/13 keys are engine-life telemetry; only 3 are config/capability constants.**

### The standout finding — alert text (`130822-alert-text-records-2026.json`, conf 94)

New record keys in the 2026 logs (`3_15`, `4_19`) decode as **alert records carrying
human-readable Mercury fault text**:

- `0x8008` (text): alertId `3181` → title **"Trim"**, message **"The trim up relay is not
  working properly."**, note **"Service engine soon."**
- `0x8018` (numeric companion): the u16s `[5, 88, 40]` are **0-based line indices into the
  Mercury fault-text catalog** (`FaultCodeTranslations/{Action,Long,Short}Text_en.txt`).

This did three things: decoded a new record type purely from logs, **named 130825 Navico
alert id 3181** (a previously-open item), and **surfaced a real boat fault** — a trim-up
relay fault present in both 2026 logs.

## What remains open (Navico) — label-only, capture-gated

The structure is complete; the only residual is the **per-key / per-field physical-quantity
NAMES**, and they are **genuinely capture-gated**. The keyed values flow through a generic
keyed model — the wire record-key is preserved verbatim and is **never bound to a static
human label** anywhere we can read it. Nor is there a public source documenting 130822's
proprietary field layout (web search 2026-06-26 returns only general SmartCraft↔NMEA 2000
gateway descriptions). So the per-key sub-quantity is decided **at runtime** when the engine
ECU populates the model, with no static binding to recover.

**Capture levers to finish it** (item `vvl_65280_130822`):

- A dated-baseline or captured **service-reset** for the engine-life keys (they are collinear
  across passive sessions, so naming needs variation).
- A **config-toggle capture** for the 3 hard-constant capability keys.
- A capture with field variation for the sub-type-2 numeric fields and the sub-type-6 TLV
  blob's internal field identities (only 1–3 static samples exist today).

## Mercury 130822 (not on our bus) — `analysis/pgn_overlay/130822_mercury-marine.json`

"Proprietary Display Status": after the 2-byte company trio (mfg 144 / industry 4), a packed
8-bit display header (`engine_inst` 4b / `helm_number` 1b / `fault_number` 2b /
`message_version` 1b) followed by five fixed-width LE data fields (u8, u8, u16, u8, u32),
12 bytes total (Fast). The first four data fields are presence-flagged against an all-ones
not-available sentinel (`0xFF`/`0xFFFF`); the u32 is always present. The five data fields are
**anonymous** (no names, no value lookups) → value semantics capture-gated. It is **binary-only
/ not present in the corpus** (no Mercury-144 130822 on our bus).

---

**Net:** 130822's wire structure is fully solved for both variants. The only residual is
naming the Navico engine-life key sub-quantities, which needs a capture (service-reset or
config-toggle) rather than further desk analysis.

PR #687, opened 2026-06-27
chore: add real-world Fusion and Victron sample captures

keesverruijt:
Real-world captures provided by @HyperActiveJ (discussion #676), as evidence for the already-merged Fusion and Victron decodes:

- **`samples/fusion-130820-status.raw`** — Fusion (mfg 419) PGN 130820, one frame per Message ID: Speed Volume Current Speed, Ignition Switch State, RDS Data, Multiroom, Multiroom Status, Processing Bypass, Mono, Menu Lock Id (#682).
- **`samples/victron-61184-vreg.raw`** — Victron (mfg 358) PGN 61184 VREG: DC Voltage (`0x2002`) and DC Current (`0x201C`) (#681).

All 10 frames decode cleanly against the current database (verified locally; no decode errors).


pete-jorgensen:
Decode is fine but there might be a bug in the json export, will send a pr.

PR #691, opened 2026-06-27
fix(json): export Signed for keyed (DYNAMIC_FIELD_VALUE) field types

pete-jorgensen:
## What

Keyed registers (`LookupFieldTypeEnumeration`, used by the Victron VREG / B&G / Simnet / Mercury key-value PGNs such as 61184 and 130824) read their integer value with the signedness of the **resolved** field type. That signedness is dropped from the JSON export.

`analyzer-explain` already emits `Signed='...'` on every `<EnumFieldType>` in `docs/canboat.xml`, and `canboat.xsd` permits the attribute, but the `canboat2json.xslt` `EnumFieldType` template never copied `@Signed` to JSON. So a decoder that works from `canboat.json` cannot distinguish a signed keyed register from an unsigned one — it sees a bare `NUMBER` either way.

## Example

`VICTRON_VREG` `0x201C` "DC Current" is typed `CURRENT_FIX32_MA` (signed). The analyzer decodes the 61184 frame `66 99 1c 20 ec fa ff ff` correctly as `-1.3 A`, but the exported entry was:

```json
{"Name":"DC Current","Value":8220,"FieldType":"NUMBER","Resolution":0.001,"Unit":"A","Bits":"32"}
```

with no `Signed`, so a JSON consumer reads `0xfffffaec` as `4294965.996 A`.

## Fix

Add `@Signed` to the `LookupFieldTypeEnumeration/EnumFieldType` template, mirroring how regular `Field`/`FieldType` records already carry `Signed`. Signed numeric registers now export `"Signed": true`, unsigned ones `"Signed": false`; lookup/string registers without sign info are unaffected. `docs/canboat.json` regenerated.

This only repairs the JSON DB consumed by external decoders — the analyzer's own decode output and all `analyzer/tests` are unchanged.


keesverruijt:
@HyperActiveJ can you PM me on Discord? You can find me in the Signal K discord, or by name.

Issue #531, opened 2025-05-06
PGN 130845 - Simnet: Key Value


Elmue:
1.)
I study this PGN and I don't understand why there is a parameter named "MinLength" ?

![Image](https://github.com/user-attachments/assets/f1006e7c-8b39-4ceb-8a3f-c059b13a8b27)

This is from logfile susteranna-actisense-serial.raw
I see that there is a communication between the controllers 0C and 0E.
Controller 0C sends a request to 0E, which is marked by a length of 0 bytes.
A few milliseconds later controller 0E sends a response with the requested value: 3009  (0x0BC1)
The response has exactly 2 bytes as indicated by the parameter "MinLength"

So my question is why is the name "MinLength" instead of "Length" ?
In the other PGNs 130824, 130833, 130846, which also contain Key-Value pairs, the length parameter is called simply "Length".
Why is it different here ?
Where is the "Minimum" here ?
I see that the exact Length count of bytes is transmitted.

2.)
I recommend to rename "Address" into "Destination Address" which makes this parameter clearer.

keesverruijt:
It's called MinLength because I also see messages where the data seems to be four bytes with the length field saying two.

I would categorise this PGN as "not fully understood yet." Until someone makes the effort to dig deeper I don't think we should change anything yet.

keesverruijt:
I checked again for #529 and the MinLength field is really a mystery right now, for this PGN the lengths are all in the lookup table it seems. Why is that minlength field there? No idea right now.

See #529 changes, and try my analyzer on `samples/simrad-ap48.raw` and notice that some lookup values are not correct yet.

Elmue:
Is it  possible that you put the name "MinLength" because you saw messages with a Length field of 0 bytes ?
I see in multiple logfiles that 0 bytes means that it is a request to another controller.

Elmue:
> It's called MinLength because I also see messages where the data seems to be four bytes with the length field saying two.

For a reason I selected the screenshot that you see above.
The response message shows exactly this situation:
```
02 : Length byte
C1 0B : Two data bytes
00 00 : Two additional bytes
```

How is this to interpret ?
If the four bytes C1 0B 00 00 would all be data bytes, then the value of the Length byte 02 would be wrong. 
It would have to be 04.

But if we think a bit deeper about it, something like a "Minimum Length" byte **cannot** even exist.
How does the receiver interpret the "minimum" length of the data ?
The receiver must always know the **exact** length to decode the data correctly.
[...]

PR #697, opened 2026-06-29
navico(130845): decode Operation byte, document composite Key

keesverruijt:
## Summary

Fixes the long-standing mystery field in PGN 130845 (**Simnet: Key Value**).

- The byte previously decoded as **`MinLength`** is not a length. Across the capture corpus it only ever takes the values **0, 1, 2**, so it is the get-set **operation** selector. It is now decoded via a new `SIMNET_KEY_OPERATION` lookup (`0 = Read`, `1 = Set`, `2 = Reply`).
- **`Key`** is widened from 2 to 3 bytes, absorbing the always-zero spare byte that followed it. It is a composite: **command group** in the low byte and the **parameter key** in the upper 16 bits.
- The two parts are **deliberately not split** into separate fields, because the parameter key is only unique *within* a command group (e.g. key `1` is "True wind high" in group 4 but "Deep water" in group 8) — the data type of `Value` can only be resolved from group and key together. This is documented on the field via a new `LOOKUP_DYNAMIC_FIELD_KEY_DESC` macro.

## Compatibility

Because the absorbed byte is always zero in observed traffic, **every existing `SIMNET_KEY_VALUE` dictionary value is numerically unchanged** — e.g. "Race timer duration" (16903) still resolves and decodes its value identically. Verified against the `samples/` corpus (3246 frames of PGN 130845).

## Field layout (after)

| # | Field | Bits |
|---|-------|------|
| Key | `DYNAMIC_FIELD_KEY` (group \| key << 8) | 24 |
| Operation | `LOOKUP` SIMNET_KEY_OPERATION | 8 |
| Value | `DYNAMIC_FIELD_VALUE` (type per key) | var |

Generated docs (`canboat.xml`/`.html`/`.json`) and golden test files regenerated in lockstep; full `make generated` (incl. test suite) passes.


keesverruijt:
Closes #531.

PR #752, opened 2026-07-08
feat(keel)!: refactor C to YAML PGN database + Rust generator + evidence-first editor

keesverruijt:
> **🚧 Work in progress — up for review, not for merge yet.**
> This PR is a large architectural refactor. It is opened as a draft so the
> **approach** can be reviewed while the remaining cleanups (below) are
> finished. Please review the direction, the data model, and the generated
> output; expect force-pushes.

## What this is

CANboat's PGN database currently *is* C source: `analyzer/pgn.h` (~10.6k
lines of a macro DSL), `analyzer/lookup.h` (~4.7k lines of X-macros included
six times), and `analyzer/fieldtype.h`. `canboat.xml` is emitted by compiling
`analyzer-explain` with `-DEXPLAIN` and hand-rolled `printf`s; XSLT then
derives `canboat.json`/`.html`.

This branch moves the source of truth to a **YAML database** under
`database/` and introduces **`keel`** (a single Rust crate, one dependency)
that validates it and generates everything downstream. The name is the idea:
like a ship's keel, the database is the thing everything else is built on.

Full rationale and decisions: **`keel/DESIGN.md`**. Contributor-facing
workflow: **`AGENTS.md` §9** and **`CONTRIBUTING.md`**.

## Why

- **The rules were invisible** — scattered across `checkPgnList()`,
  `getMinimalPgnLength()`, `fillFieldType()` and convention; you learned them
  one `logAbort` at a time, after writing C.
- **Contributing required writing C** in a macro dialect with puns
  (`unit = "=275"` meaning "match this value", `ID_AND_NAME` pinning a frozen
  id that is otherwise silently derived from the display name).
- **One 10.6k-line file** meant every pair of concurrent PRs conflicted.
- **XML emission was implicit** — presentation quirks interleaved with the
  data walk, untestable in isolation.

## What changed (a lot)

**1010 files, ~+63k/−19k.** The shape:

- **`database/`** (~931 files) — the new source of truth. One YAML file per
  PGN *variant* in `pgns/`, one per enumeration in `lookups/`, plus
  `fieldtypes.yaml`, `physicalquantities.yaml`, and a parallel `j1939/` tree.
- **`keel/`** — the Rust tool:
  - `keel check` — a real rule engine (R02–R40) replacing the scattered C
    asserts; the same rules run in CI and in the editor.
  - `keel generate` — emits `canboat.xml` **byte-identical at switchover**,
    plus the analyzer's C tables (`pgn-data.h`, `lookup.h`,
    `fieldtype-data.h`, …), all committed so building the C analyzer never
    needs the Rust toolchain.
  - `keel decode` — a from-scratch NMEA 2000 decoder (fast-packet reassembly,
    fieldtypes, dynamic key/value) used by the editor and by R40 sample
    regression tests.
  - `keel edit` — a **local, evidence-first web editor** (keel backend +
    no-build vanilla JS): paste real captures, see the bytes stacked and
    decoded live, clone a near-miss variant instead of retyping shared
    fields, edit any enumeration, adopt decodes as regression samples — all
    rule-checked before it will write. This is the intended authoring path.
- **`analyzer/`** — `pgn.h`/`lookup.h`/`fieldtype.h` are now generated; the
  `-DEXPLAIN` / v1 / macro-DSL machinery and `analyzer-explain` are deleted.
  The hand-written C keeps only the struct definitions.
- **Docs** — `AGENTS.md`, `CONTRIBUTING.md`, `README.md` rewritten so the
  documented workflow is "edit YAML / use `keel edit`", not "edit the C
  header".

Migration landed in steps (see the commit history): YAML tree + byte-identical
round-trip → rule engine → C-table generation → J1939 tree → **switchover**
(`feat!: keel owns all generation`) → decoder + editor.

## How to try it

```sh
keel/keel check          # validate the whole database
keel/keel edit           # evidence-first editor at http://127.0.0.1:8020
make generated           # regenerate every artifact + run the test suite
```

`make generated` now regenerates **before** testing and rebuilds
`canboat.html`/`json` on any database change (two ordering bugs fixed on this
branch).

## Not done yet / open for discussion

- **Pre-merge cleanups** tracked in `keel/FINDINGS.md` (e.g. unreferenced
  enumerations awaiting a keep/delete call) and the quirk keep/fix decisions
  in `keel/QUIRKS.md`.
- **Editor polish** — the wizard works end-to-end but has rough edges.
- `keel/bootstrap/` (the Python migration converter) is retained for now and
  is meant to be deleted at final merge.
- Squash strategy / how this lands (one squash vs. staged) is open.

## Review asks

1. Is the **data model** (YAML shape, lookup kinds, repeating sets) right?
2. Is the **`keel check` rule set** the right set of guarantees?
3. Does the generated `canboat.xml`/`.json` match expectations for your
   downstream consumers?
4. Is the **`keel edit`** workflow one you'd want to author PGNs with?


meaning-matters:
Why YAML; what other formats have been / could be considered?

keesverruijt:
I considered XML, and in the end decided that the actual format doesn't really matter when there are strict checks on the content in place and there is a decent editor to create new PGNs or fix existing ones (as there are in this proposal). What does matter is how easy it is to read for humans, during review. And YAML excels in that. It is also not impossible to write new YAML by hand, especially as you are only going to edit one file (PGN variant or enumeration) at a time.

Now the editor is a bit rough still, it can certainly be improved, but it doesn't need to be perfect from the start. The checks and validation must be complete though.

Note that the official "base your downstream code on this" formats remain XML and JSON, and that the output from this is byte-identical. Once this merges we can drop the byte-level compatibility and a few quirks like formatting errors in the handcrafted XML will be fixed. But if you depend on the white space in the XML, you deserve to be bitten.

tkurki:
Was going to to argue for using yaml. Used to be a yaml hater, but turned a lover, for human readability & authoring. The exact shape is probably not that important.

PR #815, opened 2026-08-05
feat: default to SI units and camelCase identifiers

dirkwa:
Implements the requested default change: the Rust binaries and libraries now default to strict SI units and camelCase (legal-identifier) names for PGNs and fields — the forms new consumers should meet first, making compatible reuse beyond Signal K more likely.

canboat C's output stays one flag away: --human restores human-readable names and --metric the practical units, on both convert and the analyzer alias. --si / --camel / --upper-camel remain accepted (now no-ops or stylistic variants), so existing invocations that pass them explicitly — signalk-server's analyzer spawn included — behave identically.

Library defaults follow suit: Units::default() is Si, JsonOptions::default() camelCase. The golden suite now pins --human --metric explicitly since its fixtures are canboat C's own outputs; the C-shape unit tests request CamelCase::Off by name, and a new test pins the Rust-wide default.

One thing deliberately left to you: I did not mark this with a conventional-commit breaking flag, to avoid release-please jumping the major — say the word and I'll amend the commit with a BREAKING CHANGE footer if that's the versioning you want.

Not changed: -nv stays opt-in, per the note that flipping it would imply changing Signal K's consumption as well.

Tested: cargo test --workspace green (26 suites) including the golden suite against the C fixtures; make rust-precommit clean.

### DIFF
--- a/crates/canboat-core/src/db.rs
+++ b/crates/canboat-core/src/db.rs
@@ -55,12 +55,13 @@ pub struct PgnDatabase {
 #[non_exhaustive]
 pub enum Units {
     /// Strict SI base units: `rad`, `K`, `Pa`, `C` (coulomb). What a
-    /// physics-facing consumer (Signal K, control code) wants. Matches
-    /// canboat C's `analyzer -si`.
+    /// physics-facing consumer (Signal K, control code) wants, and the
+    /// Rust default — machine-friendly first. Matches canboat C's
+    /// `analyzer -si`.
+    #[default]
     Si,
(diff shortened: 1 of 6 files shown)


keesverruijt:
I think this is a duplicate PR for #814?

dirkwa:
You're right — #814 covers everything here (same library defaults) plus the flag redesign and the n2kd/TUI fixes. Closing in favour of #814; I'll rebase my downstream staging on master.

PR #824, opened 2026-08-06
fix(decode): apply the range-derived sentinels to hints-less number fields

### LINKED ISSUE #823: Rust decoder: number fields with only a rangeMax miss the top-of-range sentinels the C applies
## What happens

For a number field whose schema carries a `rangeMax` but no explicit `unknown_value` / `out_of_range_value` / `reserved_value` hints, the C analyzer treats the top-of-range values as sentinels and prints `Unknown`, while the Rust decoder prints the raw reading.

Reproducer with the J1939 `TEMPERATURE_UINT8_OFFSET` fields (PGN 65262, from analyzer/tests/j1939-pgn-test.in):

```
  can0  18FEEE00   [8]  3E FF FF FF FF FF FF FF
```

```
C  : Engine Coolant Temp = 22 C; Engine Fuel Temp 1 = Unknown; ... Engine Intercooler Temp = Unknown
Rust: Engine Coolant Temp = 22 C; Engine Fuel Temp 1 = 215 C;   ... Engine Intercooler Temp = 215 C
```

0xFF is J1939's not-available pattern for these fields, so the C is right.

## Why

`extractNumberNotEmpty` (analyzer/print.c) resolves `reservedCount` for **every** number field, from `rangeMax` against the raw bit-width maximum, capped by `reservedCountForSize` — independent of whether the schema carries per-field sentinel hints.

The Rust `decode_number` only calls `sentinel_field_value`, which consults the explicit hints. The equivalent range-derived fallback exists as `range_max_sentinel` but is currently reached only from one 32-bit-specific call site (decode.rs:1715), so a hints-less field never gets the check.

## Findings from attempting a fix

Three things a fix has to get right, each of which broke a naive attempt:

1. **Offsets.** `rangeMax` is in display units, so a field with a display offset (J1939 Excess-K temperatures: raw 0 reads 233 K) needs the offset removed before comparing against the raw bit-width maximum. Otherwise 485 K looks wider than an 8-bit field can hold and the check bails out.
2. **Signedness.** For a signed field the all-ones pattern is -1, a legitimate reading the C prints as such. `range_max_sentinel` compares the *unsigned* raw against the bit-width maximum, so applying it to signed fields turns every -1 into `Unknown` (visible on `PERCENTAGE_INT8` fields in the same PGN).
3. **Unit systems.** The SI and Metric schema variants express the same field's range in different display units, so a per-field `rangeMax` authored as a Kelvin figure fixes the SI variant and leaves Metric diverging. Whatever carries the range has to be per-variant — which argues for resolving it on the field type rather than authoring it per field.

## Scope

Pre-existing: it is how offset field types without explicit hints have always behaved. The J1939 engine-temperature definitions are simply the first fixture to exercise them, so the golden test for them currently skips the affected records.

Happy to do the fix as a follow-up PR — opening this first because the third point above is really a schema-semantics decision.

dirkwa:
Closes #823.

canboat C resolves `reservedCount` for **every** number field from `rangeMax` against the raw bit-width maximum (`extractNumberNotEmpty`, print.c), whether or not the schema carries explicit per-field sentinel hints. The Rust decoder only consulted the hints, so a field carrying just a range decoded its all-ones "not available" pattern as a real reading where the C prints `Unknown`:

```
  can0  18FEEE00   [8]  3E FF FF FF FF FF FF FF      # J1939 PGN 65262

C   : Engine Coolant Temp = 22 C; Engine Fuel Temp 1 = Unknown; ...
Rust: Engine Coolant Temp = 22 C; Engine Fuel Temp 1 = 215 C;   ...
```

The range-derived check already existed as `range_max_sentinel` but was reachable only from one 32-bit call site. Two corrections were needed before it could be used generally:

- **Display shifts.** `rangeMax` is in display units, so both have to come off before comparing against the raw maximum — the schema `offset` (J1939 Excess-K: raw 0 reads 233 K) *and* the unit conversion the Metric variant applies on top. The same field's range is 485 K in SI and 209.85 C in Metric, both raw 252; dropping either term makes the range look wider than the field can hold and the check bails out. This is why a per-field `rangeMax` authored in one unit system cannot fix it — the value has to come from the per-variant schema, which keel already emits correctly.
- **Signedness.** The fallback is unsigned-only: for a signed field the all-ones pattern is -1, a legitimate reading the C prints as such, and the comparison is against the unsigned raw. (Visible on the `PERCENTAGE_INT8` fields in the same PGN.)

### Test harness deadlock

The golden harness wrote the child's stdin from the test thread while nothing drained its stdout, so any fixture whose decoded output exceeds the pipe buffer (~64 KiB) hung forever. No existing fixture was large enough to trigger it; the J1939 candump fixture is. Stdin now goes out on its own thread.

### Tested

`CANBOAT_GOLDEN=required cargo test --workspace`: 24 suites, **16/16 goldens**, zero failures — no NMEA 2000 output changes from the shared-decoder change. `make rust-precommit` clean. Verified directly against the C binary on the J1939 fixture in both unit systems.

### DIFF
--- a/crates/canboat-core/src/decode.rs
+++ b/crates/canboat-core/src/decode.rs
@@ -79,7 +79,21 @@ fn range_max_sentinel(f: &FieldInfo, ex: Extracted, bits: u32) -> Option<FieldVa
         return None;
     }
     let raw_max = (1u64 << bits) - 1;
-    let raw_range_max = (range_max / resolution + 0.5) as u64;
+    // `rangeMax` is stated in DISPLAY units, so both display shifts have
+    // to come off before it can be compared against the raw bit-width
+    // maximum: the schema's `offset` (J1939 Excess-K: raw 0 reads
+    // 233 K) and the unit conversion the Metric schema variant applies
+    // on top of it (the same field's range is 485 K in SI and
+    // 209.85 C in Metric — both raw 252). Dropping either term makes
+    // the range look wider than the field can hold, the check bails
+    // out, and the all-ones "not available" pattern decodes as a real
+    // reading where the C prints Unknown.
+    let display_offset = f.offset.map(|o| o as f64).unwrap_or(0.0);
+    let raw_range_max_f = (range_max - display_offset - f.unit_offset) / resolution + 0.5;
+    if raw_range_max_f < 0.0 {
+        return None;
+    }
+    let raw_range_max = raw_range_max_f as u64;
     if raw_range_max >= raw_max {
         return None;
     }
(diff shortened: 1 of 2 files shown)


dirkwa:
Merged as part of #825, which was stacked on this branch — 5696fca is on master and #823 is closed. Nothing left to rebase; closing.

PR #855, opened 2026-08-25
ci(rust): actually run the MSRV job on the MSRV toolchain

keesverruijt:
The `Rust MSRV` job resolves `rust-version` from `Cargo.toml` and installs that toolchain, but then ran a bare `cargo check`. `dtolnay/rust-toolchain` selects a toolchain with `rustup default`, and **that loses to the repo's `rust-toolchain.toml`**, which pins `channel = "stable"`.

rustup says so itself, in the job log:

```
info: note that the toolchain 'stable-x86_64-unknown-linux-gnu' is currently in use
      (overridden by '/home/runner/work/canboat/canboat/rust-toolchain.toml')
```

So the job has been re-checking **stable** and reporting it as an MSRV pass. It has never been able to catch an MSRV regression — the green badge was meaningless.

Setting `RUSTUP_TOOLCHAIN` does not help either; the toolchain file wins over the environment variable too. `rustup run <version>` names the toolchain explicitly and does take precedence.

## The change

```diff
-        run: cargo check --workspace --locked
+        run: rustup run ${{ steps.msrv.outputs.version }} cargo check --workspace --all-targets --locked
```

Also widened to `--all-targets` so test and bench code is held to the MSRV as well — that's the "extend it to actually test MSRV" half.

## Verification

Checked against a **real** 1.88 toolchain locally (via `rustup run`, since the toolchain file otherwise silently redirects to stable):

- `cargo check --workspace --locked` → passes
- `cargo check --workspace --all-targets --locked` → passes

So making the job genuine does not turn it red; 1.88 is an accurate floor today.

## Ordering

This branch is cut from `master`, so its **`Rust Clippy` job will fail** until #854 lands — that failure is the pre-existing clippy 1.98 breakage, not anything in this PR. Happy to rebase once #854 merges.

## Summary by CodeRabbit

* **Chores**
  * Updated the minimum supported Rust version to 1.96.
  * Improved continuous integration checks to consistently validate the workspace against the configured minimum supported Rust version.
  * Ensured locked dependency versions are used during compatibility checks.

### DIFF
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -99,8 +99,12 @@ jobs:
       - name: Cache cargo & target
         uses: Swatinem/rust-cache@v2
 
+      # `rustup default` (set by dtolnay/rust-toolchain) loses to the repo's
+      # rust-toolchain.toml, which pins `stable` - so a plain `cargo check`
+      # here silently re-tests stable instead of the MSRV. `rustup run` names
+      # the toolchain explicitly and wins over the file.
       - name: Check
-        run: cargo check --workspace --locked
+        run: rustup run ${{ steps.msrv.outputs.version }} cargo check --workspace --all-targets --locked
 
   clippy:
     name: Rust Clippy
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ edition = "2024"
 license = "Apache-2.0"
 authors = ["Kees Verruijt <kees@verruijt.net>"]
 repository = "https://github.com/canboat/canboat"
-rust-version = "1.88"
+rust-version = "1.96"
 
 [workspace.dependencies]
 canboat-schema = { path = "crates/canboat-schema", version = "8.0.0-beta3" } # x-release-please-version


keesverruijt:
Update: with the job actually running on the toolchain it advertises, it immediately caught a real problem — **the declared MSRV of 1.88 was not true**:

```
error: rustc 1.88.0 is not supported by the following packages:
  vergen@10.0.2       requires rustc 1.96.0
  vergen-gitcl@10.0.2 requires rustc 1.96.0
  vergen-lib@10.0.2   requires rustc 1.96.0
```

`crates/canboat-bridge/build.rs` uses `vergen-gitcl = "10"` to stamp git version info, and the 10.0.2/10.0.3 line has a 1.96 floor. The options were to pin back to `vergen-gitcl 9.1.0` (the newest release still on 1.88) or to raise the declared MSRV to match reality. Going with the latter — `rust-version = "1.96"`, pushed as a second commit.

Worth noting *why* this stayed hidden even from local testing: a warm `target/` dir will happily build the workspace on 1.88, because cargo only rejects the graph when it actually **resolves** — which CI does on every clean checkout. So this was never going to show up in normal local work.

Verified against a real 1.96 toolchain (`rustup run 1.96`, with `rust-toolchain.toml` moved aside so nothing silently redirects to stable): `cargo check --workspace --all-targets --locked` passes. Nothing else in the tree pinned 1.88.

PR #883, opened 2026-09-18
feat(fusion): add samples, fix spellings and normalise label casing

keesverruijt:
## What does this PR do?

Follow-up cleanup for the four Fusion PRs that just landed — #878, #879, #880 and #881.

### Samples

Those PRs contributed real captures but only #879 added `samples:` expectations, so the new decodes were not pinned by anything. This adds them, generated from the contributed captures and verified against the analyzer:

| PGN | Coverage |
|---|---|
| 126720 Tuner Command | all five commands across both AM and FM, including the injected Tune Direct at 101.7 MHz |
| 126720 Set Aux Gain | +9, 0, -9 dB, plus the -10 dB command the MS-RA70N clamps |
| 130820 Aux Gain | +9, 0, -9 dB — the status side, which never reports beyond ±9 |
| 130820 Balance | -7, 0, +7, pinning the signed decode |
| 130820 Low Pass Filter | all four enumerated cutoffs |

The Aux Gain pair is worth a look: the capture shows the command side accepting -10 and +10 while the status side only ever reports -9 and +9, which is exactly the clamping #880 described.

**Sublevels and Volume Limits are deliberately left unsampled.** Both carry an 8-byte fast-packet payload, and that cannot be expressed in a single `raw:` line. Keel reads a raw of 8 bytes or fewer as one CAN frame — so it wants a fast-packet header, which pushes the line to 10 bytes — and reads anything longer as a coalesced message, so that header is then decoded as payload. Neither round-trips. Every existing sample in the database happens to be longer than 8 bytes, which is why this has not come up before.

### Capture format header

`samples/fusion-ms-ra70n-zone-controls.raw` had no format header. Without one the analyzer guesses `PLAIN_OR_FAST` and rejects every frame:

```
ERROR [analyzer] Received incomplete fast packet PGN 130820 from source 12
```

With `# format=FAST` all 29 messages decode. As merged, the capture could not be replayed by canboat's own tools.

### Spelling

Three `FUSION_MESSAGE_ID` labels were misspelled:

- 16: `DAB Updtate Command` → `DAB Update Command`
- 19: `Set Low Pass Filer` → `Set Low Pass Filter`
- 52: `Scan Multirooom Devices` → `Scan Multiroom Devices`

`tools/contract-pr.sh` rates a relabel a minor contract change, hence the `feat:` type.

### Also

126720 Set Aux Gain now cites its capture in `notes:`, as the other three variants do.

## Not included

Left for a decision rather than changed unilaterally, since each is a contract change:

- casing inconsistencies — `SiriusXm Signal` vs `SiriusXM *` elsewhere, `Ipod` vs `iPod`, `Send Dvd Command` / `Dvd Press Ir Key` vs `DVD`, `Fapi` vs `FAPI`, and the lowercase `USB repeat` / `iPod shuffle` / `DVD region` entries in `FUSION_SETTING`
- `130820 Balance` field is named `VALUE` in caps, unlike every neighbouring field
- the zone-control capture also contains 126720 command frames for proprietary IDs 18, 19, 20 and 23 (Set Balance / Low Pass Filter / Sublevel / Volume Limit) that have no variant yet and currently fall through to the 0x1EF00 fallback

## Checklist

- [x] PR **title** follows Conventional Commits
- [x] Edited the **YAML under `database/`** and `keel/keel check` passes (13 pre-existing string-padding errors, unchanged from master)
- [x] Ran `make generated` and committed the regenerated files in the same commit
- [x] Ran `make pr` — reports `minor`, declared `minor`, OK
- [x] New decodes are backed by real captures already in `samples/`

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected command, source, status, tuner, and setting labels for spelling, capitalization, and consistency.
  * Standardized display names for DAB Update, Low Pass Filter, DVD, DSP, SiriusXM, iPod, and USB features.
  * Corrected Fusion balance and menu-lock terminology.

* **Improvements**
  * Expanded validation for auxiliary gain, balance, and low-pass filter values.
  * Added tuner samples covering FM and AM seek, tuning, and direct frequency selection.
  * Added coverage for negative, zero, and positive audio adjustment values.


lmjogback:
Sorry for the extra work and the missing samples: expectations.
This was my first PRs, and I'll try to improve in the future.

I have already prepared a PR that depends on #881 
- Adds Set Balance, using a signed 8-bit value.
- Adds Set Low Pass Filter, using the observed 55, 85, 120, and 160 Hz lookup values.
- Adds Set Sublevel.
- Adds Set Volume Limit.
- Adds Set Zone Name using STRING_LZ.
- Corrects the existing `Set Low Pass Filer` typo to `Set Low Pass Filter`.
- Adds capture evidence including a Zone 1 rename to `COCKPIT` and the corresponding status response.

I'll rebase that and submit it for review if that is ok?

lmjogback:
I did create #884 on master for the Set commands mentioned above.

keesverruijt:
Extended with the full casing audit (second commit).

The audit was mechanical: tokenise every label in the `FUSION_*` lookups plus every Fusion PGN description and field name, and flag any token spelled more than one way **within the Fusion namespace**. An earlier pass compared against the whole database and was useless — lowercase words are perfectly correct mid-sentence in other manufacturers' labels.

That found 15 relabels, plus one thing the scan could not classify: four of the 49 `FUSION_MESSAGE_ID` entries — and only those four — carried a `Fapi ` prefix, an internal Fusion API namespace leaking into what is otherwise a flat verb-first list. Dropping the prefix rather than recasing it puts them beside their neighbours (`Set Source` 2, `Get Multiroom Device Record` 51) and collides with no existing label.

Also caught: `130820 fusionMarineTuner`'s description was the literal string `Fusion: MARINE_TUNER` — the raw enumeration name, where every other Fusion PGN has prose.

Worth noting that the `130820 fusionSettings` samples pinned four of the old labels and failed R40 the moment the rename landed. That is the regression check doing exactly what it exists for, and it is the argument for the samples added in the first commit.

After this, no token is spelled two ways anywhere in the Fusion lookups, descriptions or field names. `tools/contract-pr.sh` rates the 20 relabels **minor** and the two description changes **cosmetic**, so it all fits in this release rather than needing another one.

keesverruijt:
@lmjogback no worries, good that someone is finally picking up all this Fusion stuff and working through it. I will open an issue about "Fusion: clean up PGNs" and then we can use that to converse about these changes. The PRs need to be focused.

In future, add "proof" samples to the YAML files. This keeps the decoders stable and reliable over time!

PR #884, opened 2026-09-18
feat(fusion): add zone control commands

lmjogback:
## What does this PR do?

Adds five Fusion zone-control command variants for proprietary PGN 126720:

* **Set Balance** — signed 8-bit balance value; negative values shift left, positive values shift right, and zero is centered.
* **Set Low Pass Filter** — decodes the observed 55, 85, 120, and 160 Hz filter settings using `FUSION_LOW_PASS_FILTER`.
* **Set Sublevel** — decodes the zone and subwoofer level.
* **Set Volume Limit** — decodes the zone and volume-limit value.
* **Set Zone Name** — decodes the zone name using `STRING_LZ`.

The commands were reverse engineered and verified using NMEA 2000 traffic between a Fusion MS-NRX300 remote and an MS-RA70N stereo.

The variants use `variantOrder` 60 through 64, following the recently merged Tuner Command and Set Aux Gain variants at 58 and 59.

## Capture and sample coverage

The existing `samples/fusion-ms-ra70n-zone-controls.raw` capture provides evidence for all five variants. It includes:

* Balance values of -7, 0, and +7.
* All four observed low-pass filter settings.
* Sublevel values of 13 and 8.
* Volume-limit commands for both exposed zones.
* A Zone 1 rename to `COCKPIT`, followed by the corresponding status response.

Each new PGN definition includes `samples:` expectations derived from the capture.

The six-byte command payloads are represented as eight-byte CAN frames: the `00,06` fast-packet header followed by the six-byte payload. The longer Set Zone Name message is represented as a coalesced payload.

## Relationship to #883

This PR is based on current `master`, which already includes #878, #879, #880, and #881.

The open cleanup PR #883 touches the same lookup and capture files. The only semantic overlap is the correction of `Set Low Pass Filer` to `Set Low Pass Filter`; the capture additions in this PR are different. If #883 merges first, I will rebase this branch and drop the duplicated spelling fix and regenerated output.

Feedback is very welcome. I'm happy to revise the PR, rebase it, or adjust its scope if that would make review and merging easier.

Created with the support of ChatGPT

## Checklist

* [x] PR **title** follows Conventional Commits (`feat` / `fix` / … — see [CONTRIBUTING.md](../CONTRIBUTING.md))
* [x] If I changed the PGN database, I edited the **YAML under `database/`** (not the generated files) and `keel/keel check` passes
* [x] I ran `make generated` and committed the regenerated files in the same commit
* [x] Tests pass (`make tests`)
* [x] I ran `make pr` and acted on its contract report
* [x] New or changed decodes are backed by a real capture and include `samples:` expectations

## Summary by CodeRabbit

- **New Features**
  - Added support for decoding additional Fusion Electronics marine audio messages.
  - Added PGN definitions for zone balance, low-pass filter, sublevel, volume limit, and zone name settings.
  - Added decoding for zone identifiers, audio values, filter frequencies, and zone names.
  - Included documentation and sample messages covering supported zones and setting values.


keesverruijt:
Reviewed this, rebased it onto current `master` locally and tested it. It is in good shape: `keel check` stays at the pre-existing 13-error baseline, `make generated` (which runs `make tests`) is clean, `variantOrder` 60–64 is free, and all 15 of the `126720` frames in the capture decode where every one of them previously fell through to the 0x1EF00 fallback. The sample encoding is right too — the `00,06` fast-packet framing for the six-byte payloads and coalesced form for the 14-byte Set Zone Name is exactly what keel's sample decoder wants.

One thing to add before merge, and then some independent confirmation.

## The zone field is zero-based, and nothing says so

All five new variants have a `zone: UINT8`, and in every sample it is `0`. It is **zero-based**, while the fixed four-slot status messages and Fusion's factory default names are one-based. Someone writing an encoder will set `zone=1` meaning "Zone 1" and hit the wrong zone.

The clean evidence is already in your own capture — Sublevel, where only one slot moves:

```
10:09:56  cmd  Set Sublevel zone=0 level=0d(13)  ->  status  Zone 1: 13, Zone 2: 12
10:10:07  cmd  Set Sublevel zone=0 level=08      ->  status  Zone 1:  8, Zone 2: 12
```

`zone=0` moves the slot canboat calls **Zone 1**, and Zone 2 never moves.

Could I ask for a line in `notes:` on all five saying the zone is zero-based? That is the whole fix.

## Confirmed on a different Fusion model

I had access to a **Fusion MS-WB670** (SW 6.50, build 121) on a live bus, driven by a B&G Zeus SRX over NMEA 2000 — a different model from your MS-RA70N, so this is cross-model confirmation rather than a second sample of the same unit. I asked the owner to name the zones and then change one by name. Their zone names:

```
Zone Name index 0 = "Doghouse"    (the zone the UI calls Zone 1)
Zone Name index 1 = "Saloon"      (the zone the UI calls Zone 2)
Zone Name index 2 = "Zone 3"      (still on its factory default name)
```

Note index 2 defaulting to the name `"Zone 3"` — the same corroboration your existing `130820 fusionZoneName` sample shows with `number: 2` → `"Zone 3"`, now reproduced on different hardware.

They then set the balance of *Saloon* to left 7. The Zeus put this on the bus:

```
14:31:10.253  cmd     a3,99,12,00, 01,f9     proprietaryId 0x12 = 18 = Set Balance
                                   |  +- f9 = -7
                                   +- zone = 1
14:31:10.264  status  Balance      zone=1  value=-7
```

That is the first non-zero-zone command frame from any controller — every sample in this PR and every frame in your capture is `zone=0` — and **your `fusionSetBalance` definition decodes it correctly**, zone and signed value both, on a model it was not written from. It also independently confirms the balance direction from #881: negative is left, in the operator's own words.

## Smaller: one concept, two field names

`fusionSetZoneName` calls it `zone`; the existing `130820 fusionZoneName` status calls the same thing `number`. Since #883 just normalised Fusion naming, worth aligning — though moving the status side to `zone` is a contract change, so probably not something to fold in here silently.

## Conflicts

#883 has landed since you opened this, so this was showing as conflicted. The conflicts were only in generated files and resolved by regenerating; your duplicate `Set Low Pass Filer` -> `Set Low Pass Filter` fix converged on the same line, so there was nothing to hand-resolve. I have pushed the merge to your branch so you do not have to redo it — your database YAML is untouched, only the regenerated output, and CI is green.

## Related, but not for this PR

On the MS-WB670 the Zeus sends **proprietary id 21** for sublevels — a value absent from `FUSION_MESSAGE_ID` — which writes all four zones at once, where your NRX300 sends **20** `Set Sublevel` for a single zone. Both are real; different controllers take different paths. So `fusionSetSublevel` here is correct, just not the only form. I will put 21 in a separate PR rather than complicate this one.

Thanks for these, and for the captures. They made all of the above checkable rather than a matter of opinion.

lmjogback:
Thanks for the thorough review, the independent verification, and for handling the merge!

I've added the zero-based zone numbering clarification to `notes:` in all five definitions and regenerated the artifacts in f4ac034f. `make pr` completed successfully.

I agree that aligning `number` with `zone` on the existing status definition is best handled separately, since that changes the public contract.

PR #887, opened 2026-09-18
feat(fusion): decode Set Equalizer, Set All Sublevels, Loudness and Zone Capabilities Extended

keesverruijt:
## What does this PR do?

Four new decodes from a live NMEA 2000 bus, plus one field rename for consistency.

All of it comes from a **Fusion MS-WB670** (SW 6.50, build 121) driven by a **B&G Zeus SRX**, captured 2026-09-18. Every decode was identified by changing a named control and watching the wire, so the names come from what the operator read off the screen rather than from guessing at the byte layout.

## New PGN 126720 variants

**Set Equalizer (proprietary id 22)** — the command form of the 130820 `Tone` status. Same payload shape, and every command carries the whole bass/mid/treble triple rather than only the axis that moved:

```
14:24:38.768  cmd     a3,99,16,00, 00,08,03,00  ->  status  zone=0 bass=8 mid=3 treble=0
14:25:27.912  cmd     a3,99,16,00, 00,00,03,00  ->  status  zone=0 bass=0 mid=3 treble=0
14:25:30.186  cmd     a3,99,16,00, 00,00,00,00  ->  status  zone=0 bass=0 mid=0 treble=0
```

**Set All Sublevels (proprietary id 21)** — writes all four zone subwoofer levels at once, mirroring the 130820 `Sublevels` status, the way 25 `Set All Volumes` mirrors 24 `Set Zone Volume`. Id 21 was **missing from `FUSION_MESSAGE_ID`**, which jumped 20 → 22.

Both forms are real and are sent by different controllers: the Zeus SRX sends **21**, while the MS-NRX300 wired remote in #884 sends **20** for a single zone. So `fusionSetSublevel` in that PR is correct, just not the only form.

## New PGN 130820 variants

**Loudness (message id 32831)** — a per-zone flag, and the id was **missing from `FUSION_STATUS_MESSAGE_ID`**, which jumped 32830 → 32832. Identified by flipping the control on two different zones in two separate sessions; each flip emitted a single unsolicited frame for that zone alone while the periodic sweep reported the others unchanged.

**Zone Capabilities Extended (message id 32865)** — named in the enumeration but never decoded. Five 8-byte records: four zone slots plus a global slot. The framing is confirmed against a PGN canboat already decodes rather than guessed — the leading word of each record is exactly the corresponding word of 32798 `Capabilities`, which the same unit sent in the same window:

| | Capabilities (32798) | 32865 record |
|---|---|---|
| Zone 1 | 0xFDFF | 0xFDFF |
| Zone 2 | 0xBDFF | 0xBDFF |
| Zone 3 | 0xB7FF | 0xB7FF |
| Zone 4 | 0x0000 | 0x0000 |
| Global | 0x0000 | 0x0000 |

The six trailing bytes of each record read `02,00,00,00,00,00` for populated slots and zero otherwise. **Their meaning is not established, so they are left as `BINARY`** rather than given speculative names.

## One breaking change: `number` -> `zone`

`130820 fusionZoneName` called its zone selector `number`. It selects a zone, and every other zone-selecting field in the Fusion PGNs is called `zone` — including the 126720 `Set Zone Name` command in #884 that writes this very value.

The six `130568`..`130575` entertainment PGNs also have a `number` field, but theirs is documented as *"Source number per type"* — a source index, not a zone — so those are deliberately left alone.

## Two variants have no `samples:` block

`fusionSetEqualizer` and `fusionSetAllSublevels` both have an **exactly 8-byte payload**, which cannot be expressed in a single `raw:` line. Keel reads a raw of 8 bytes or fewer as one CAN frame — so it wants the fast-packet header, pushing the line to 10 bytes — and reads anything longer as a coalesced message, so that header is decoded as the manufacturer word and the wrong variant is selected. Neither round-trips.

This already leaves `130820` Sublevels and Volume Limits unsampled (#883), so it now blocks four PGNs across two PRs. Their captures are included regardless, and both decode correctly through the analyzer.

## Evidence

Every capture carries a `# format=FAST` header and a comment block stating what was done and what is *not* established.

| File | Frames | Decodes |
|---|---|---|
| `fusion-set-equalizer.raw` | 10 | all |
| `fusion-set-sublevels-all-zones.raw` | 2 | all |
| `fusion-loudness.raw` | 14 | all |
| `fusion-zone-capabilities-extended.raw` | 3 | all |
| `fusion-wb670-zone-indexing.raw` | 18 | 9 (the Set Balance commands need #884) |
| `fusion-tuner-region.raw` | 5 | all |
| `fusion-volume-zone-sync.raw` | 3 | all |

The last two are evidence for settings semantics that are documented in `notes:` but not yet enumerated, because canboat has no per-setting value enumeration for `FUSION_SETTING` today:

- **Tuner Region** (id 7): `0` USA, `1` Europe, `2` Japan, `3` Australasia — every value observed against a name on screen. A region change also clears all 30 AM/FM presets **unconditionally**, one `Setting` message per preset, even when they are already empty. That is the same mechanism #879 documented for "Delete All", reached by an unrelated trigger, which rules out its being an artifact of that menu action.
- **Volume Zone Sync** (id 74): a zero-based zone index with a sentinel — `0` and `1` select a zone, `4` means All. Not a bitmask; a four-zone mask would be 15. What `4` derives from is not established and is not claimed.

## Checklist

- [x] PR **title** follows Conventional Commits
- [x] Edited the **YAML under `database/`** and `keel/keel check` passes (13 pre-existing string-padding errors, unchanged from master)
- [x] Ran `make generated` and committed the regenerated files in the same commit
- [x] Ran `make pr` — reports `minor` for the rename, declared `feat:`
- [x] New decodes are backed by real captures added to `samples/`

## Summary by CodeRabbit

* **New Features**
  * Added support for Fusion commands to set all subwoofer levels and adjust equalizer settings.
  * Added decoding for Fusion loudness status and extended zone capability messages.
  * Added lookup values for new Fusion message and status identifiers.
* **Improvements**
  * Clarified Fusion zone naming by using “zone” consistently in Zone Name data and examples.
  * Added sample coverage and decoding details for newly supported Fusion messages.


keesverruijt:
Rebased onto master now that #884 and #885 have landed. Conflicts were only in generated output and resolved by regenerating; the database YAML is unchanged.

Two things that improved as a result:

**The zone-indexing capture now decodes completely.** When I opened this, 9 of its 18 frames fell through to the 0x1EF00 fallback because `fusionSetBalance` did not exist yet. With #884 merged:

```
  3x  Fusion: Zone Name
  9x  Fusion: Set Balance
  6x  Fusion: Balance
```

Those nine are the first non-zero-zone command frames captured from any controller, and they decode through a variant written from a different Fusion model with only zone-0 samples. I have updated the capture's header accordingly — it previously said the frames would not decode until #884 landed.

**The `number` -> `zone` rename now closes a gap #884 opened.** That PR added `fusionSetZoneName` with a `zone` field, writing the very value the `fusionZoneName` status reported as `number`. Command and status now agree.

`variantOrder` 65 and 66 slot in cleanly after #884's 60–64, with no shadowing anywhere in the database.

PR #890, opened 2026-09-19
feat(fusion): decode setting values via an indirect lookup, starting with Tuner Region

### LINKED ISSUE #888: Fusion: FUSION_SETTING 76 is mislabelled, and several observed message ids remain unidentified
Loose ends from a live-bus session on a **Fusion MS-WB670** (SW 6.50, build 121) with a **B&G Zeus SRX**, 2026-09-18. The decodes that were conclusive landed in #887; these are the ones that need more evidence before anything is changed.

## 1. `FUSION_SETTING` 76 is not "BT Auto Connect"

The operator changed a setting the Fusion app calls **"Power On Volume Limit"** to 24. Setting id **76** ramped smoothly `13 → 24` over ~750 ms of slider dragging:

```
15:08:43.407  id 76   13 -> 14
       ...
15:08:44.154  id 76   23 -> 24
```

A Bluetooth auto-connect setting does not have a draggable 0–24 range. A volume limit does, and 24 is exactly the ceiling seen throughout — `Volume Limits` read `24,24,24` and no zone volume exceeded 24. Note that `75 Max Volume Start`, the obviously matching name, never moved.

**This is not general drift in the upper ids.** Three were confirmed correct in the same session: `7 Tuner Region`, `67 Internal Amps On` and `74 Volume Zone Sync` all moved exactly when their named control was changed.

So either 76 is an isolated mislabel, or the list is shifted by one from 75 onward — in which case the real `BT Auto Connect` sits at **77** (currently `Null Setting`). **The test that distinguishes them:** flip a Bluetooth auto-connect toggle and see which id moves. Landing on 77 means shifted; anywhere else means an isolated error.

Worth settling before renaming, since a label change is a minor contract break and the fix differs completely between the two cases.

## 2. Message ids observed but absent from `FUSION_STATUS_MESSAGE_ID`

Four ids are broadcast by the MS-WB670 and are in no enumeration, so they decode as raw fallback bytes today. Two hours of flipping every control the Zeus and the iOS app expose never moved any of them, and three have only ever shown a single value:

| id | shape | observed |
|---|---|---|
| `0x805C` (32860) | zone + byte, all three zones | always `0` |
| `0x805D` (32861) | zone + byte, **only zones 0 and 1** | always `0` |
| `0x8071` (32881) | single byte, global | always `1` |
| `0x8072` (32882) | single byte, global | always `0` |

`0x805D` skipping zone 2 is the strongest lead — zone 2 is a line-level pre-out on this installation while zones 0 and 1 drive speakers, so it is something only an output-capable zone has. Candidates not ruled out: subwoofer enable, speaker size / crossover, zone enable, telemute.

Identifying these needs whichever control sets them, which may be one no MFD or app exposes — the same wall hit with `Set Loud`.

## 3. Commands still unreachable

`FUSION_MESSAGE_ID` **54 `Set Loud`** and **34 `Set Zone Name`** with a non-zero zone could not be captured. Neither the Zeus SRX nor the Fusion iOS app exposes loudness or zone renaming over NMEA 2000 — and in their default configuration both drive the head unit over **Ethernet**, putting nothing on CAN at all. In a 25-minute session with both active, the bus saw exactly two `126720` frames, both `Request Status` from the Zeus at boot.

Disabling Ethernet on the Zeus moved it onto NMEA 2000, which is how #887's commands were captured at all. A wired remote (MS-NRX300) is likely to expose more of the head unit's DSP menu than an MFD does.

## 4. `Capabilities` (32798) is static, not live state

Verified rather than assumed: with `Internal Amps On` switched **off**, a `Request Status` forced a fresh copy and every byte was identical, including bit 14 of the Zone 1 word.

```
             setting 67   Z1        bit 14
15:17:51     amp ON       0xFDFF    set
15:22:07     amp OFF      0xFDFF    set
```

So a client cannot use `Capabilities` to know whether a zone will currently produce sound. Two further properties worth documenting somewhere: it is broadcast on a **5-minute** cycle, much slower than the ~50-second zone sweep, and it is also answered on demand via `Request Status` within ~2 seconds.

Separately, bit 14 (`0x4000`) is set for exactly the one zone that has an **internal** amplifier and clear for the externally-amplified one. That is consistent across the three zones of a single installation and fits a static-capability reading, but it is one device and is not claimed as decoded.

## 5. A sample cannot be written for a Fast PGN with a 7–8 byte payload

A payload of exactly 8 bytes cannot be expressed in a single `raw:` line. Keel reads a raw of 8 bytes or fewer as one CAN frame, so it wants the fast-packet header, which pushes the line to 10 bytes and is then read as a coalesced message with that header decoded as the manufacturer word.

This now blocks four PGNs: `130820` Sublevels and Volume Limits (#883), and `126720` Set Equalizer and Set All Sublevels (#887). It has not surfaced before because every other sample in the database happens to be longer than 8 bytes.

## 6. `FUSION_SETTING` values have no enumeration

#887 documents two value mappings in `notes:` because there is nowhere better to put them:

- id 7 `Tuner Region`: 0 USA, 1 Europe, 2 Japan, 3 Australasia
- id 74 `Volume Zone Sync`: zero-based zone index, 4 = All

canboat supports `INDIRECT_LOOKUP` with a `triplet` enumeration keyed on another field's value — the mechanism `060928 isoAddressClaim` uses for `DEVICE_FUNCTION` keyed on device class. That is the natural home for these, keyed on the setting id, but it changes the `value` field type on `fusionSetting`, `fusionSettings` and `fusionSetSetting` and so wants doing deliberately rather than folded into a decode PR.

keesverruijt:
Closes #888 item 6.

`FUSION_SETTING` names the settings but nothing described their **values**, so `Tuner Region = 1` was as far as canboat got. #887 documented the mapping in `notes:`, which never reaches `canboat.json` — so no consumer could see it.

canboat already had the mechanism: `INDIRECT_LOOKUP` with a `triplet` enumeration keyed on another field's value, exactly as `060928 isoAddressClaim` uses for `DEVICE_FUNCTION` keyed on device class.

| PGN | field | key |
|---|---|---|
| `130820` fusionSetting | `value` | field 5 (`id`) |
| `126720` fusionSetSetting | `value` | field 5 (`id`) |

New `FUSION_SETTING_VALUE` enumeration, currently covering id 7 `Tuner Region`: **0 USA, 1 Europe, 2 Japan, 3 Australasia** — every value observed against a name read off the unit (`samples/fusion-tuner-region.raw`, from #887).

## It degrades the way it has to

A mapped value decodes to its name; an unmapped one decodes as the plain number it always did. That was the main risk — most Fusion setting values are frequencies and levels, not enumerable. Verified:

```
ID='Tuner Region'      Value='Japan'         <- mapped
ID='Volume Zone Sync'  Value=4               <- not enumerated
ID='FM Preset 0'       Value=102500000       <- not enumerated
```

#879's preset-frequency samples pass unchanged.

## Two things this turned up

**The XSD was internally inconsistent.** `LookupEnumeration/@MaxValue` is `xs:long`, but `LookupIndirectEnumeration/@MaxValue` was `xs:int` — so a 32-bit indirect enumeration failed validation on `4294967295` while the 32-bit *direct* `FUSION_SETTING` lookup had always been fine:

```
canboat.xml:4173: element LookupIndirectEnumeration, attribute 'MaxValue':
  '4294967295' is not a valid value of the atomic type 'xs:int'
```

Widened to `xs:long` to match. Widening an attribute's type is backward compatible for readers.

**The repeating `130820 fusionSettings` is deliberately NOT converted.** I tried it first, and it is actively unsafe. The indirect key is resolved by `extractNumberByOrder`, which computes a fixed bit offset from the start of the message — so inside a repeating set it reads the **first record's** id for every record:

```
Tuner Region   1            <- with the lookup applied, still 1: the key read was record 1's id
```

That is not merely ineffective. Had the first record's setting happened to be enumerated, every later record would have been labelled with *that* setting's value names. Tried, observed, reverted, and the reason recorded in that PGN's `notes:` so it does not get "fixed" back.

## What is deliberately left out

Id **74 `Volume Zone Sync`** is not enumerated. Its value is a zero-based *zone index* with a sentinel — 0 and 1 select a zone, 4 means All — so the meaningful values depend on the unit's zone count rather than forming a closed list. It stays documented in `notes:`.

## Contract

Minor: `value` changes from `NUMBER` to `INDIRECT_LOOKUP` on two PGNs, plus the new enumeration. `keel check` stays at the 13-error baseline.

## Summary by CodeRabbit

- **New Features**
  - Fusion setting values can now display descriptive region names for USA, Europe, Japan, and Australasia.
  - Tuner Region values are automatically decoded using these named regions.
  - Fusion Set Setting values support lookup-based decoding where applicable.

- **Documentation**
  - Added guidance explaining why repeated fusion settings continue to use numeric values for reliable decoding.

- **Improvements**
  - Expanded the supported range for lookup values in the XML schema.


keesverruijt:
Filed #891 for the underlying decoder issue this turned up.

Short version: the reason `130820 fusionSettings` does not get the indirect lookup here is not Fusion-specific. The C analyzer and canboat-rs both locate an `INDIRECT_LOOKUP`'s key field by a statically computed bit offset, which is only correct for its first occurrence, so inside a repeating set they read the first record's key for every record. keel's decoder uses a runtime `by_order` map and gets it right — so the same database decodes differently depending on which tool reads it.

It is latent today: all four `INDIRECT_LOOKUP` fields in the database, including the two added here, are in non-repeating PGNs. Once #891 is fixed, `fusionSettings` can have the lookup and the note explaining its absence can go.

PR #892, opened 2026-09-19
feat(decode): resolve an INDIRECT_LOOKUP key by position, fixing repeating sets

### LINKED ISSUE #891: INDIRECT_LOOKUP: all three decoders resolve the key field differently, and each is wrong on a different shape
An `INDIRECT_LOOKUP` field takes its meaning from a second field, named by `lookupIndirect.order`. Two of canboat's three decoders locate that key field by a **statically computed bit offset**, which is only correct for the field's first occurrence. Inside a repeating set they read the **first record's** key for every record.

keel's decoder does it correctly, so the same database decodes differently depending on which tool reads it.

## Current state

| decoder | how it locates the key | repeating sets |
|---|---|---|
| **keel** — `keel decode`, R40, the editor | runtime `by_order` map, overwritten as each field decodes (`keel/src/decode.rs`) | **correct** |
| **C analyzer** | `getFieldOffsetByOrder()` sums field sizes from the start of the message (`analyzer/print.c:136`, used by `extractNumberByOrder` at `:321`, called from `fieldPrintLookup` at `:733`) | **wrong** |
| **canboat-rs** | `val1_field.bit_offset`, the precomputed static offset (`crates/canboat-core/src/decode.rs:1456`) | **wrong** |

keel walks the repeating set by re-running `decode_one` over the same field objects, and each one does `ctx.by_order.insert(f.order, e.value)` before the lookup is resolved — so record *n*'s key is record *n*'s value. The other two have no equivalent notion of "the current record".

## Observed

Applying an indirect lookup to the repeating `130820 fusionSettings` (each record is `{id, value}`, key = the record's `id`):

- **keel** resolves each record against its own `id`
- **the C analyzer** resolves every record against the *first* record's `id`, so all of them miss the enumeration and print as bare numbers

This is not merely ineffective. Had the first record's key happened to be enumerated, **every later record would have been labelled with that key's value names** — a silent mis-decode rather than a missing one.

## Severity today: latent

Nothing in the database currently triggers it. Every `INDIRECT_LOOKUP` field lives in a non-repeating PGN:

```
 60928 isoAddressClaim      deviceFunction  key=order 7   (no repeating set)
 65240 isoCommandedAddress  deviceFunction  key=order 7   (no repeating set)
126720 fusionSetSetting     value           key=order 5   (no repeating set)
130820 fusionSetting        value           key=order 5   (no repeating set)
```

The only present cost is that `130820 fusionSettings` cannot have the lookup its single-setting siblings got in #890, which is documented in that PGN's `notes:` so it is not "fixed" back by someone who notices the asymmetry.

## Shape of the fix

Both broken decoders can derive the key's position from where the current field *actually* is, rather than where it statically sits:

```
delta  = staticOffset(valueField) - staticOffset(keyField)
keyAbs = actualOffset(valueField) - delta
```

In C the running absolute offset is already in hand as `startBit`, and `Field.order` exists, so it is a handful of lines. canboat-rs is the same shape, using the `bit_offset` already passed to `decode_indirect_lookup`.

## The trap

**That form is only valid when the key and the value are in the same repeating set.** If the key is in the fixed part and the value repeats, the relative form yields `K + n*S` where the answer is `K` — it would break the case that works today.

Both existing users, `isoAddressClaim` and `isoCommandedAddress`, are exactly that fixed-part-key case. So the fix must branch on whether the key's order falls inside the same repeating block, and **the regression risk is on the working path, not the broken one**.

## Suggested approach

1. Fix the C analyzer and canboat-rs **together** — fixing one alone deepens the divergence.
2. Guard on whether the key order lies within the same repeating set; fall back to the static offset otherwise.
3. Add coverage for **both** cases — a key in the fixed part (as the ISO PGNs have) and a key in the same repeating set — since only the second is currently broken and only the first is currently exercised.
4. Once all three agree, give `130820 fusionSettings` its indirect lookup and drop the note explaining why it does not have one.

keesverruijt:
Closes #891.

An `INDIRECT_LOOKUP` field takes its meaning from a second field named by `lookupIndirect.order`. All three decoders located that key differently, and **each was wrong on a different shape**:

| decoder | mechanism | key **before** value (repeating `{id, value}`) | key **after** value (`60928`) |
|---|---|---|---|
| C analyzer, canboat-rs | static bit offset | wrong — reads record 1 every time | correct |
| keel | decode-order map | correct | wrong — key not yet seen |

The C/Rust failure is not merely ineffective: had the first record's key been enumerated, **every later record would have been labelled with that key's value names**.

keel's failure was live, not theoretical — `60928 isoAddressClaim` has a forward reference (`deviceFunction` is order 5, keyed on `deviceClass` at order 7), and keel decoded it as a bare `130` where the C analyzer resolved `Atmospheric`.

## The rule, now shared by all three

```
delta = staticOffset(value) - staticOffset(key)
keyAt = actualOffset(value) - delta
```

applied **only when key and value share a repeating set**; otherwise the static offset is used, which is correct whether the key precedes or follows the value, because such a key occurs exactly once.

That guard is the important part: both existing users (`isoAddressClaim`, `isoCommandedAddress`) are the fixed-part-key case that already worked, so the regression risk sat on the working path.

## Verified

Same PGN 130820 `Settings` capture through all three decoders — each record is `{id, value}`, keyed on the record's own `id`:

```
Language           0              not enumerated -> stays numeric
Volume Zone Sync   4              not enumerated -> stays numeric
Tuner Region       'Europe'       resolved, per record
FM Preset 0        102500000      not enumerated -> stays numeric
```

And the forward reference, unchanged in C, now correct in keel:

```
with the fix:     decoded Atmospheric (130)
without the fix:  decoded 130
```

R40 accepts a resolved lookup's numeric raw as matching, so no existing sample expectations needed changing.

## Payoff

`130820 fusionSettings` gets the indirect lookup its single-setting siblings received in #890, and the note explaining why it couldn't have one is removed.

## Checks

`keel check` 0 errors, `make generated` (incl. `make tests`) clean, `cargo test` green, `cargo fmt --check` and `clippy` clean. Contract: minor, from `fusionSettings.value` becoming `INDIRECT_LOOKUP`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed indirect lookup decoding for values and keys within repeating field sets.
  - Repeating records now resolve their corresponding keys instead of incorrectly using the first record’s key.
  - Preserved correct handling for keys outside repeating sets, including forward references.
  - Updated Fusion settings message decoding to support indirect lookup values with 32-bit keys.


keesverruijt:
Note on the checks quoted in the description: `keel check` reports **0 errors, 0 warnings** on master and with this change.

Earlier PRs in this series cited a "13-error pre-existing baseline" — that was wrong, and #888 has the correction. keel is a workspace member so its binary lands in `target/release/keel`, but I had been running `keel/target/release/keel`, a leftover from before that move, dated 2026-07-26. The `keel/keel` wrapper handles this correctly; I bypassed it. The stale `keel/target/` has been removed.

Nothing merged was affected — everything re-verifies at 0 errors — but the numbers quoted as evidence in #883, #887, #889 and #890 came from a two-month-old checker.

PR #895, opened 2026-09-19
feat(fusion): decode External and Internal Amp Gain (130820 message ids 32860/32861)

keesverruijt:
Closes part of #888 item 2.

Two message ids a Fusion MS-WB670 broadcasts that were in **no enumeration**, so they decoded as raw fallback bytes:

```
32860  External Amp Gain
32861  Internal Amp Gain
```

Both are `{zone, gain}`. Zone is zero-based; gain is **signed**, the same shape as `130820` Aux Gain.

## What identifies the pair

An asymmetry visible on the bus. The unit has three zone slots, the third being a **line-level pre-out with no internal amplifier**:

- `32860` is broadcast for **all three** zones
- `32861` only for the **two that have an internal amplifier**

Two hours of toggling every control the stereo's app and a B&G Zeus SRX expose never moved either value — both sat at 0 throughout — so the bus alone could not name them. The subwoofer-amplifier toggle, my best hypothesis, turned out to be a local-only setting never broadcast at all.

## Where the names and layout come from

Garmin's Fusion Audio Android app (`fusion.link.remote` 13.0.0). Its message enumeration names them `FAPI_EXT_AMP_GAIN` and `FAPI_INT_AMP_GAIN`, and their decoders read:

```java
zone = getUInt8(buffer, 2);
gain = getInt8 (buffer, 3);
```

The same app's decoder for `32831` Loudness reads zone as `UInt8` and the flag as `UInt8`, independently confirming the variant added in #887.

## A limitation, stated plainly

**Every captured frame has gain 0.** The signedness rests on the app's decoder, not on an observed negative value. That's recorded in both PGNs' `notes:` rather than presented as though the captures proved it — if someone later captures a non-zero gain, it is worth checking against this.

`keel check` 0 errors, no variant shadowing. Contract: additive.

## Summary by CodeRabbit

* **New Features**
  * Added decoding support for Fusion external amplifier gain messages.
  * Added decoding support for Fusion internal amplifier gain messages.
  * Gain data is now available by audio zone, including signed gain values.
  * Added recognition for the corresponding Fusion status message identifiers.
  * Improved per-zone tracking for Fusion amplifier gain, loudness, and mono data.


keesverruijt:
Good catch, and it went further than the two files flagged.

Checked the convention across every per-zone `130820` message. Five already key on `zone` — Balance, EQ, Line Level Control, Low Pass Filter, Zone Name. **Four did not**: the two new ones here, plus `fusionLoudness` (added in #887, so mine) and `fusionMono` (pre-existing). All four are fixed in c055086.

So the same defect has been live for Loudness since #887 and for Mono since it was first defined — a snapshot would have collapsed all zones into a single record for both. Worth having found that by checking the whole convention rather than just patching what was flagged.

The two pre-existing fixes make this PR minor rather than purely additive; the contract check reports `fusionLoudness` and `fusionMono` `zone primaryKey: False -> True`.

PR #898, opened 2026-09-20
test(rust): raise coverage from 57% to 71%, add a coverage job, fix a SocketCAN load bug

keesverruijt:
## Why

There was no coverage measurement anywhere in the repo, so nobody knew the number. It was **56.8%** line coverage across the Rust workspace (554 tests). This takes it to **71.3%** (702 tests on macOS, 704 on Linux with vcan) and adds a CI job so it stays visible.

## Coverage

Measured with `cargo llvm-cov --workspace --all-features`.

| crate | before | after |
|---|---:|---:|
| canboat-schema | 100.0% | 100.0% |
| canboat-cli | 91.5% | 91.5% |
| canboat-io | 74.2% | 81.2% |
| canboat-core | 77.9% | 79.7% |
| canboat-tokio | 73.6% | 73.6% |
| **keel** | **11.6%** | **68.6%** |
| **canboat-bridge** | **53.0%** | **65.8%** |
| canboat | 58.0% | 58.0% |
| **total** | **56.8%** | **71.3%** |

`crates/canboat/` is unchanged — what is left there is `tui/ui.rs` rendering and `main`-adjacent wiring, which wants a different kind of test than this PR is doing.

## The measurement was wrong on macOS

`canboat-io/src/device/socketcan.rs` is entirely inside `#[cfg(target_os = "linux")]`. On a macOS dev box only the 23-line unsupported stub compiles, so the file reported 0% — and the four tests already living in `mod imp` had never been compiled, let alone run. Everything in this PR that touches that file was written and verified on Linux (aarch64, rustc 1.98.1).

With a vcan interface available, that file goes from an apparent 0% to **75.6%**.

## The bug this found

`compute_load_pct` carried a comment claiming its `wrapping_sub` kept a counter wrap from showing up as phantom traffic. It did the opposite: `read_sysfs_counter` parsed the kernel's u64 and truncated it `as u32`, and a `wrapping_sub` at **u64** width cannot undo a wrap that happened at **u32** width. Across a 4 GiB `rx_bytes`/`tx_bytes` boundary the delta came out as ~1.8e19 instead of a few hundred bytes, saturated, and published a spurious **100% CAN bus load** in PGN 262400.

The load sampler now reads the counters at their native u64 width, so there is no wrap to undo. The u32 wrapper stays for `rx_errors` / `tx_dropped`, whose PGN fields are u32.

**This is the one behaviour change in the PR** — everything else is tests and CI. Easy to drop if you would rather keep it as a known issue; the test that catches it is `a_counter_past_four_gigabytes_is_not_a_wrap`.

## What the tests cover

- **`device/maretron.rs`** (0% → covered) — the reception-driven handshake, where the flip to binary is triggered by `CONNECTED` arriving rather than by the writer.
- **`device/socketcan.rs`** — load maths, ISO NAME packing, fast-packet chunking at the 223-byte ceiling; plus three tests that drive a **real socket** on vcan and so enter the worker thread, poll loop, `recv_batch`, `handle_frame`, `dispatch_cmd` and the TX ring.
- **`analyze.rs`** — every `Config` knob, none of which had a test of its own.
- **`garmin_csv` / `chetco` / `actisense_ascii` / `airmar`** — the error paths a real capture hits, plus the fast-packet clamp.
- **`n2kd/decoded.rs`** (no test module at all) — every 0183 sentence, with the NMEA checksum verified on each.
- **`n2kd/ais_decoded.rs`** — rather than snapshotting opaque armoured payloads, these unarmour the payload back to bits and assert fields at their ITU-R M.1371 offsets, so a wrong bit width shows up as a wrong MMSI. Types 1/4/5/9/12/14/18/19/21/24, plus fragment splitting.
- **`keel/tests/database.rs`** — runs `keel check` and `keel generate --check` against the real `database/` tree inside `cargo test`. This is most of keel's jump. It also means a stale committed artifact now fails `cargo test`, not just CI — the runtime compiles `schema_generated.rs`, so that mattered. Runs in 0.6s.

## CI

A `coverage` job with `--fail-under-lines 65` — a regression floor, not a target, sitting ~6 points below today so ordinary work has headroom. Linux only, for the `cfg` reason above. It also writes the per-file table into the job summary.

A `setup-vcan` composite action brings up `vcan0` for the coverage job and for the Linux leg of the test matrix. It **fails** when vcan is unavailable rather than tolerating it: the tests no-op when the interface is missing, so a tolerant step would silently drop that coverage with every job still green.

⚠️ Whether `modprobe vcan` works on a GitHub-hosted runner is the one thing I could not verify locally — runners are full VMs rather than containers, and vcan is in the stock Ubuntu module set, so it should. **If the "Set up vcan" step is the only thing red, that is the answer and I will relax it to tolerate a missing interface.** The vcan tests themselves were verified on real hardware.

## Verification

- macOS: 702 tests, 0 failures, clippy clean, `cargo fmt` clean.
- Linux (aarch64): 704 tests, 0 failures, including all 25 SocketCAN tests against a live vcan0, stable across 5 consecutive runs.

## Summary by CodeRabbit

- **New Features**
  - Added support for parsing Airmar capture-format data.
  - Added automated code coverage reporting and virtual CAN testing.

- **Bug Fixes**
  - Preserved large SocketCAN counter values during load sampling.
  - Improved load calculations after interface counter resets.
  - Improved handling of malformed input and unavailable CAN interfaces.

- **Quality Improvements**
  - Expanded validation for AIS, NMEA 0183, Actisense, Garmin, Maretron, and other data-processing workflows.
  - Added broader SocketCAN frame exchange and fast-packet coverage.
  - Added validation for generated database artifacts and deterministic output.


keesverruijt:
Both findings from the earlier review are addressed in 25298aa:

- **`compute_load_pct` counter reset** — agreed, and it was the same defect as the u32 truncation this PR already fixed, just reached by a different route. It now returns `None` for that interval so PGN 262400 carries canboat's sentinel instead of a flat 100%. Verified the premise first: `prev_load_sample` advances unconditionally at socketcan.rs:672, so a reset costs one sample, not the series — pinned by a second test.
- **`assert_ne!(n2k, j1939)` passing on `None`** — correct, a regression that stopped J1939 decoding entirely would have left it green. Added `assert!(j1939.is_some(), …)` before the comparison.

The previous review covered 054f43c; the branch is now at 25298aa, two commits further on.

@coderabbitai full review

PR #901, opened 2026-09-20
feat(fusion): decode RDS station names, via a new per-field encoding: declaration

lmjogback:
## What does this PR do?

Document an RDS character-encoding bug observed on at least the Fusion MS-RA70N in the notes for PGN 130820, Fusion Tuner.

The station name for Radio Bohuslän is sent in RDS G0 ([EBU Latin](https://en.wikipedia.org/wiki/EBU_Latin)) encoding without conversion to UTF-8. The captured bytes for `Bohuslän` are `42 6f 68 75 73 6c 91 6e`: `0x91` represents `ä` in RDS G0, whereas UTF-8 would use `c3 a4`. I observe the same incorrect character on the radio's own display and in the decoded CAN output.

Add eight original candump frames from my MS-RA70N/MS-NRX300 capture, covering the alternating eight-character names `Bohuslän` and ` Radio  `, and cite the sample file from the PGN notes.

This documents the observed device bug; it does not change field definitions or decoder behavior, or assume the same encoding for other models or text fields.

Validation completed in my environment: `keel/keel check` passed with 0 errors and 0 warnings; `make generated`, `make tests`, and `make pr` passed. The contract report found no contract changes. Regenerated files are committed, including the EBU Latin reference update in `863de2f8`.

Created with the support of ChatGPT

## Checklist

- [x] PR **title** follows Conventional Commits (`feat` / `fix` / … — see [CONTRIBUTING.md](../CONTRIBUTING.md))
- [x] If I changed the PGN database, I edited the **YAML under `database/`** (not the generated files) and `keel/keel check` passes
- [x] I ran `make generated` and committed the regenerated files in the same commit
- [x] I ran `make pr` and acted on its contract report (mark the title `type!:` if it reports a breaking change)
- [x] New or changed decodes are backed by a real capture (added to `samples/`) or public documentation

## Summary by CodeRabbit

* **New Features**
  * Added support for decoding non-UTF-8 Fusion tuner text using the Basic RDS character set.
  * Added optional encoding metadata to field definitions and generated schema outputs.
  * Added validation for supported encodings and eligible text fields.

* **Bug Fixes**
  * Corrected malformed UTF-8 handling so Radio Bohuslän displays “ä” correctly instead of an invalid character.

* **Documentation**
  * Added technical notes covering MS-RA70N behavior, RDS2/DAB handling, applicable standards, and supporting evidence.
  * Updated the schema version to 2.6.0.


keesverruijt:
Addressed in 3ceceb4. You were right that the schema annotation still described `EBU_LATIN` and ETSI TS 101 756 V1.8.1 — that was the spelling and the table this branch started with, before it settled on `RDS_G0` and V1.6.1. The module docs, the YAML note and both generated headers were updated in that switch; `docs/canboat.xsd` was not.

Worth spelling out why that one mattered most: the schema is the contract third-party readers implement against, so it named a value keel rejects (R41) *and* pointed at the DAB mapping this mechanism exists to avoid, while explicitly steering readers away from the correct table. The two tables agree on every accented letter common in a station name, so it would not have shown up in testing.

Also added `the_schema_documents_exactly_these_encodings`, which reads `docs/canboat.xsd` with `include_str!` and asserts keel's accepted charsets are all named there and that the discarded name is not — the same approach `rules.rs` uses to keep its rule inventory in step with `check.rs`. Mutation-checked by reintroducing the old spelling.

The previous review covered 33df61f; the branch is now at 3ceceb4.

@coderabbitai full review

keesverruijt:
@coderabbitai Thanks — you found something real, but not the thing described, and it is wider than this PR.

**The finding as stated does not reproduce.** `case '\377':` cannot match `c == 255` on a signed-`char` platform, because `'\377'` is `-1` there. On macOS ARM64 and x86-64 Linux that branch is dead code, so the RDS mapping is reached and nothing truncates:

```
frame: …,c7,03,91,ff,41,00   (Tuner Track = 0x91 0xFF 'A', declares RDS_G0)
  C (signed char):   'äÿA'
  Rust:              'äÿA'
```

**What is actually wrong is worse.** The same source compiled with `-funsigned-char` — the default on ARM Linux, which is what `build-armv7-musl` and `build-keel-musl (aarch64-…-musl)` produce — *does* match, and truncates:

```
  C (unsigned char): 'ä'
```

So the C analyzer's 0xFF handling is platform-dependent, and canboat's CI builds both families.

**And it is not specific to `RDS_G0`, or to this PR.** The same split appears on a field that declares no encoding at all, so it predates the change:

```
frame: …,126998,…,05,01,41,ff,42,…   (STRING_LAU 'A' 0xFF 'B', no encoding:)
  C (signed char):   'AÿB'
  C (unsigned char): 'A'
  Rust:              'AÿB'
```

Declining the proposed diff: special-casing `rds_g0` inside `case '\377'` would fix one field type on one platform family and leave the general bug in place, while making the inconsistency harder to see.

The real fix is `case 0xff:` so the intent holds on every platform, plus a decision on which semantics win — the comment says 0xFF means end-of-field, Rust instead maps it and relies on the trailing trim. That changes output for every string field on every device and touches golden files, so it wants its own PR rather than riding along with the encoding mechanism. Happy to open an issue with these repros.

Trailing 0xFF is unaffected either way: `printString` trims it before the escaper ever sees it, which is why no existing test catches this.

keesverruijt:
Filed as #902 with both repros, rather than expanding this PR.

